PR #66, opened 2026-06-17
Add CloudPanel Varnish purge support to the agent

BoostLabsAU:
## What
Adds CloudPanel-aware cache purging to the WPMgr agent. On CloudPanel sites the agent detects Varnish from the site user's `~/.varnish-cache/settings.json` and clears CloudPanel Varnish alongside WPMgr's disk page-cache purges:

- full-site purges send a host PURGE and a cache-tag PURGE,
- per-URL purges clear the matching Varnish URLs (path and query preserved),
- full-site purges also clean the host PageSpeed cache when it is writable.

The optional CloudPanel WordPress plugin is not required. CloudPanel joins the existing host
integrations (Varnish, Cloudflare, Kinsta).

## Why
CloudPanel operators can run WPMgr page cache alongside CloudPanel Varnish without installing
the CloudPanel WordPress plugin or manually clearing two cache layers after every WPMgr purge.

## Changes
- New `CloudPanel` integration class, registered in `Integrations`.
- Focused PHPUnit coverage in `IntegrationsPurgeTest`.
- Operator docs in `docs/agent.md` and `docs/features/caching.md`, plus a CHANGELOG entry.

## Checklist
- [ ] `make lint` passes
- [ ] `make test` passes
- [ ] `make build` passes
- [ ] OpenAPI updated (if endpoints changed) — N/A
- [x] Docs updated (if behavior changed)
- [x] Touches auth / crypto / agent protocol / RBAC / tenant isolation? → No (cache-purge behavior only)

## Summary by CodeRabbit

* **New Features**
  * Added CloudPanel Varnish cache purging support to the agent. Full-site purges clear local disk cache and trigger host-scoped and cache-tag `PURGE` requests to CloudPanel Varnish, and wipe the CloudPanel host PageSpeed cache when writable; public varnish endpoints are ignored. Per-URL purges send targeted `PURGE` requests (up to 200) preserving path/query and skipping offsite URLs.
* **Documentation**
  * Updated the agent and caching guides with CloudPanel Varnish detection and purge/PageSpeed cleanup behavior, noting the optional CloudPanel WordPress plugin is not required.


mosamlife (commented):
Thanks for this — it's a high-quality PR: the URL-count cap, the realpath + containment + symlink guards around the PageSpeed deletion, and the per-value header sanitization are all well done, and the test coverage is real. I traced the call paths and the `$urls` reaching `purgeUrls()` come from WPMgr-internal sources (`get_permalink`/`get_term_link`), not request bodies, which lowers the blast radius. No critical or high findings, and your "no auth/crypto/protocol" self-assessment is accurate. Two **medium** items to address before merge, plus a couple of nits.

**1. The "loopback-only" claim isn't actually enforced (`class-cloudpanel.php`).** The class docblock and the `purge()` comment state the target is CloudPanel's local loopback endpoint, but `normalizeEndpoint()` accepts *any* `http(s)://host[:port]` from `settings.json` — there's no loopback/private-range check — while `purge()` sets `reject_unsafe_urls => false` and `sslverify => false`. The connection target comes from the operator-owned `~/.varnish-cache/settings.json` (not request input), so this isn't remotely exploitable, but the code doesn't match its own security claim, and `reject_unsafe_urls => false` is only justified *if* loopback is actually enforced. Please either (a) enforce loopback/private-range in `normalizeEndpoint()` (reject public hosts) so the comment is true, or (b) drop the "loopback" wording and document that the endpoint is operator-trusted config. The sibling `Varnish` integration hard-pins `127.0.0.1` for reference.

**2. Varnish over-purge via an attacker-controllable Host (`class-cloudpanel.php`).** `purgeUrls()` sets the `Host` header from each URL's parsed host, and `purgeAll()` builds it from `$_SERVER['HTTP_HOST']` (visitor-controllable). On a shared Varnish fronting multiple vhosts, a spoofed `Host` could purge a *neighbor* tenant's cache (cache-eviction DoS, not data disclosure). Please validate the host against the current site (e.g. derive from `home_url()` / skip URLs whose host isn't this site) rather than trusting the request/URL host. Preferring `home_url()` over `HTTP_HOST` in `siteHost()` also tightens this.

**Nits (optional):**
- A one-line comment acknowledging the realpath→unlink TOCTOU window in `deleteDirectoryContents()` would be good (the containment checks make it low-risk, but it's worth noting).
- Add a negative test asserting a non-loopback `server` value is rejected, once #1 is addressed.

The PageSpeed deletion itself is safe — I confirmed the `realpath` + double `pathInside` containment + `is_link`/symlink handling hold, so nothing can delete outside `pagespeed_cache/v3/<host>`. Once #1 and #2 are in, this is good to merge. Nice work.

BoostLabsAU:
I've addressed both medium findings and the nits.

On Medium #1 (loopback enforcement): We went with option (a) and enforced it in the code. Added an isPrivateAddress() helper that checks if the endpoint is loopback (127.0.0.1 or ::1) or in the private ranges (10.x, 172.16-31.x, 192.168.x). normalizeEndpoint() now calls this and rejects anything public with a clear error message. Code now ctually matches what the docblock claims.

On Medium #2 (Host-header spoofing): Changed siteHost() to derive from home_url() instead of the request Host header. Then in purgeUrls() we validate each URL's host against the configured site host and skip any that don't match. Mismatched hosts never get sent to Varnish with the wrong Host header. So even if a visitor tries to spoof the Host, we only purge our own site's cache.

On the nits: Added the TOCTOU comment in deleteDirectoryContents() explaining the small window between realpath() and unlink() and why the containment checks make it safe. Also added test_cloudpanel_rejects_public_varnish_endpoint() which asserts that trying to configure a public endpoint like example.com gets rejected.

mosamlife (commented):
Thanks for the security pass — I re-reviewed the updated code and both findings are genuinely closed. Nice work.

**Security: approved.**
- **Loopback/private-range enforcement (M1):** `normalizeEndpoint()` now requires a literal private/loopback IPv4 (`filter_var(FILTER_VALIDATE_IP, IPV4)` + the 10/8, 172.16-31/12, 192.168/16 ranges, plus `127.0.0.1`/`::1`), and fails closed by throwing on anything else. Because no hostname can pass, there's no resolver step and therefore no DNS-rebinding path, so `reject_unsafe_urls=false` / `sslverify=false` are acceptable here. Integer/octal/IPv4-mapped IPs and credential/port/path smuggling are all rejected too. Good.
- **Host header (M2):** the outbound `Host` is now derived from `home_url()` (server config), each URL's host is validated against the site host via `sameHost()` and skipped on mismatch, and `MAX_URLS` bounds the loop. The tests that spoof `HTTP_HOST=attacker.example` and assert the purge still targets the site host are exactly the right coverage.
- PageSpeed deletion containment (realpath + double `pathInside` + symlink guards) is intact, and the agent test suite passes in CI.

**One non-blocking robustness item (your call):** `normalizeEndpoint()` throwing `InvalidArgumentException` on a public/misconfigured endpoint propagates an uncaught fatal through the `wpmgr_purge_*` hook dispatch, so an operator misconfiguration turns into a fatal on every purge. Consider failing soft instead — return empty/null and no-op the purge (matching the other invalid-settings cases) — and have `test_cloudpanel_rejects_public_varnish_endpoint` assert "no HTTP call + no throw". A one-line comment in `isPrivateAddress()` noting that only the `127.0.0.1`/`::1` literals (not all of `127.0.0.0/8`) are accepted by design would also help future maintainers.

**Before this can merge:** the branch now conflicts with `main` (we shipped several releases today, 0.51.2 to 0.51.5, that touch `CHANGELOG.md` and the agent). Could you rebase / merge the latest `main` and resolve the conflicts? CI is green; once it's conflict-free (and ideally with the fail-soft tweak above) this is good to go from my side.

BoostLabsAU:
I’ve pushed the follow-up at `8322f85`: public/non-private CloudPanel Varnish endpoints still fail closed, but now normalize to no usable endpoint and the purge hooks no-op instead of throwing. The regression test now dispatches the purge hook and asserts no exception path, no HTTP calls, and no PageSpeed deletion for a public endpoint.

I also added the small `isPrivateAddress()` comment clarifying that only exact `127.0.0.1` / `::1` loopback literals are accepted by design, not the whole `127.0.0.0/8` range.

The branch has also been refreshed against latest `main` and the `CHANGELOG.md` conflict is resolved.

mosamlife:
Thanks for the updates and the security hardening (loopback/private-range enforcement + fail-soft on public endpoints) — the code looks good and all four CI checks are green. ✅

The only thing blocking merge is a conflict with `main`, and it's **purely in `CHANGELOG.md`** — there are **no code conflicts**. `main` advanced a lot today (the `0.53.0` → `0.57.0` security-suite + vulnerability-scanner releases all landed), so the changelog entries collide.

To unblock, please merge the latest `main` into this branch and resolve `CHANGELOG.md` by **keeping both sides**:

```
## [Unreleased]

### Added
- **CloudPanel cache purge support in the agent.** …   ← your entry stays

## [0.57.0] - 2026-06-21
…                                                       ← main's new sections stay
## [0.54.0] - 2026-06-20
```

i.e. keep your `[Unreleased]` CloudPanel entry on top, then keep main's `## [0.57.0]` … `## [0.54.0]` version sections below it. Once you push that, CI will re-run and I'll merge. 🙏

PR #69, opened 2026-06-19
Fix scoped backup cache excludes

BoostLabsAU:
## What

Fix backup archiving so default runtime and staging excludes are scoped to known wp-content roots instead of matching any directory segment named `cache`, `upgrade`, or `upgrade-temp-backup`.

- Preserve plugin and theme vendor code in paths such as
`plugins/builderius/vendor-prefixed/symfony/cache/Adapter/ArrayAdapter.php`
- Continue excluding runtime cache and update-staging roots such as `cache`
`uploads/cache`, `upgrade`, and `upgrade-temp-backup`
- Add focused regression coverage for both `files.list` and plugin archive
  contents.

## Why

The previous default exclude list matched any path segment named `cache`. That correctly skipped runtime cache folders, but it also skipped legitimate Composer package source under plugins. A completed snapshot could therefore restore with required plugin files missing.

This keeps cache artifacts out of backups while preserving plugin and theme source trees that happen to contain directories with those generic names.

Verification note: branch CI passed on commit
`8632d5fcefe172c81806ee6739c4cf3135b74915` in run `27799248661` (`PHP (agent)`, `Go (apps/api)`, `Security audit`, and `JS (web)`).

## Checklist

- [ ] `make lint` passes (branch CI passed)
- [ ] `make test` passes (branch CI passed)
- [ ] `make build` passes (branch CI passed)
- [x] OpenAPI updated (if endpoints changed) - N/A, no endpoint changes
- [x] Docs updated (if behavior changed) - `CHANGELOG.md`
- [x] Touches auth / crypto / agent protocol / RBAC / tenant isolation? -> No

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved backup exclusions to preserve plugin and theme vendor code within cache and update-related directories (`cache`, `upgrade`, `upgrade-temp-backup`), while keeping runtime cache and update staging roots excluded as intended.
* **Tests**
  * Added regression coverage to verify backup include/exclude behavior for cache, uploads cache (excluding cache while archiving dated uploads), and update staging scenarios, including the updated default exclusion rules.


mosamlife (commented):
Thanks for this — it's a real correctness fix and a good catch. I traced it against current `main` and it holds up:

- **Root cause is right.** The old `DEFAULT_EXCLUDES` matched any path *segment* named `cache` / `upgrade` / `upgrade-temp-backup`, so legitimate plugin/theme vendor trees (e.g. `plugins/.../vendor-prefixed/symfony/cache/...`) were silently dropped — a restore could come back missing required files. Splitting into segment-excludes (scratch + secret filenames) vs anchored `DEFAULT_PATH_EXCLUDES` (runtime/staging roots) is the correct shape.
- **The `sourceDir == wp-content` assumption holds.** `FilesArchiver` is only constructed with `wpContentPath()` (`includes/backup/class-task-runner.php:534`), so `cache`, `uploads/cache`, `upgrade`, and `upgrade-temp-backup` resolve to the intended wp-content roots. `CoreFilesArchiver` is separate and unaffected.
- **Secrets stay excluded.** `wpmgr-object-cache-config.php` / `.wpmgr-oc-state.json` remain in the segment list — good, those must never enter an archive.
- **Tests are solid.** They assert against both `files.list` and the actual zip entries, and cover the now-preserved vendor dir, the still-excluded roots, and the constant split.

Two things to confirm before merge (not blockers, mainly so the intent is documented):

1. **Please document the deliberate narrowing.** With anchored prefixes, only top-level `wp-content/cache` and `wp-content/uploads/cache` are skipped now; a runtime cache nested elsewhere (e.g. `uploads/<year>/cache`, or a plugin writing transient cache under `uploads/<plugin>/cache`) that the old broad segment match used to drop will now be **included** in the backup. That's the safe direction (better to back up a cache file than lose source), and the major page-cache plugins (WP Rocket / W3TC / Autoptimize) all live under `wp-content/cache/...` so they stay excluded. Could you add a one-line comment on `DEFAULT_PATH_EXCLUDES` noting this trade-off is intentional, so a future reader doesn't "re-broaden" it back to a segment match and reintroduce the bug?

2. **Caller `exclude_paths` staying segment-based** (per your inline comment) — just confirming that's intended. It preserves existing operator-configured behaviour, so it's fine.

Once #1 has a short doc comment, this looks good to merge from my side. Confirmed no auth/crypto/protocol surface, as you noted. Nice fix.

BoostLabsAU:
Thanks for the thorough review. 

I've added the doc comment on DEFAULT_PATH_EXCLUDES to document the intentional narrowing and explain why nested plugin cache is included. The comment makes clear this is a deliberate trade-off and warns future readers not to re-broaden it back to bare segment matching.

On item #2: Confirmed, exclude_paths staying segment-based is intentional. We narrowed DEFAULT_EXCLUDES (WPMgr-hardcoded) from bare segments to anchored roots to fix the vendor-code bug, but kept operator-configured exclude_paths segment-based to preserve backward compatibility. Operators who configured segment exclusions like ['node_modules', 'backup', 'cache'] expect WPMgr to exclude those folder names anywhere on the site, not just at root. Forcing anchored matching on operator config would break existing setups. The segment-based approach also gives operators flexibility.

Commit 2ed08e24 pushed to fix/backup-cache-excludes with the doc comment. Both items addressed and ready when you are.

mosamlife (commented):
Thanks — the `DEFAULT_PATH_EXCLUDES` doc comment is exactly what I was after, and the fix itself looks good. One mechanical thing before merge: the branch now conflicts with `main` (we shipped a few releases today, 0.51.2 to 0.51.5, that touch `CHANGELOG.md` and the agent). Could you rebase / merge the latest `main` and resolve the conflicts? CI is green; once it's conflict-free this is ready to merge from my side.

mosamlife:
Thanks for keeping this rebased. Same situation as #66: the only merge conflict with `main` is in **`CHANGELOG.md`** — **no code conflicts**. `main` moved a lot today (the `0.53.0` → `0.57.0` releases), so the changelog entries collide.

Two things to unblock:

1. Merge the latest `main` and resolve `CHANGELOG.md` by **keeping both** your entry and main's new `## [0.57.0]` … `## [0.54.0]` sections (keep your bullet under `## [Unreleased]`).
2. The required CI checks (**Go (apps/api) / JS (web) / Security audit / PHP (agent)**) don't appear to have run on the latest commit — `main` is branch-protected and needs all four green. Pushing the rebase should trigger them.

Once it's conflict-free and CI is green, I'll merge. 🙏

PR #93, opened 2026-06-21
fix(api): isolate media encoder river schema

BoostLabsAU:
## What

Run the media-encoder's River client in its own Postgres schema so it no longer
competes with the API for River leadership.

A new optional setting, `WPMGR_RIVER_MEDIA_SCHEMA`, controls this:

- Empty (default): unchanged single-schema behavior.
- Set (for example `media_encoder`): the API keeps its River client on the default
  schema for all API workers and periodics, and adds a second insert-only client
  (no workers, no periodics, never started) that enqueues `media_encode` and
  `site_screenshot_capture` into the media schema. The media-encoder runs its River
  client on that same schema and is the only leader there.

Supporting changes:

- A small `riverutil` helper validates the schema identifier, qualifies raw
  `river_job` references, and idempotently creates/migrates the media schema and
  grants privileges to the app role using the migration-owner pool.
- The waker and the encoder drain path count `river_job` against the configured
  schema and the encoder-owned queues instead of always `public.river_job`.
- `.env.example`, the bundled compose profile, and the install / media-optimizer
  docs set and explain the schema, including that the same value must be set on
  both the API and the encoder.

The bundled self-host compose profile sets `media_encoder`, so self-hosters get the
fix by default.

## Why

The API and the media-encoder both start River clients on the same schema. River
elects one leader per schema to run the periodic job enqueuer. When the encoder wins
that election it only knows its own periodic jobs, so the API's `uptime_probe` and
`cron_kick` jobs stop being scheduled and uptime monitoring goes silent while the
encoder is running. Giving the encoder its own schema lets both run together: the API
stays leader of the default schema (uptime and cron keep scheduling) and the encoder
leads the media schema (image optimization and screenshots keep running).

River v0.38.0's `Config.Schema` and `rivermigrate.Config.Schema` make this possible;
the schema and app-role privileges are created explicitly at startup.

Deployment notes:

- Set the same `WPMGR_RIVER_MEDIA_SCHEMA` on the API and the encoder. Mismatched
  values can strand media jobs.
- Empty or `public` preserves current behavior.
- Existing `public.river_job` rows are not migrated. Deploy during a quiet window or
  drain media jobs first.

## Checklist

- [x] `make lint` passes
- [x] `make test` passes
- [x] `make build` passes
- [ ] OpenAPI updated (if endpoints changed): n/a, no endpoint or contract changes
- [x] Docs updated (if behavior changed)
- [ ] Touches auth / crypto / agent protocol / RBAC / tenant isolation: no. The only
  privilege surface is granting the app role usage and River table/sequence
  privileges on the new schema, mirroring the existing River grants.

Testing: added unit coverage for the new config/schema validation (default, `public`,
custom, invalid) and for the schema-aware waker and drain counters, plus a dual-schema
integration test asserting API-default, media, and screenshot enqueues land in the
expected schema. The encoder package and that integration test sit outside the default
`ci` job's package set (the CGO encoder image and container-based integration tests),
so they were built and run separately and pass.

## Summary by CodeRabbit

* **New Features**
  * Added `WPMGR_RIVER_MEDIA_SCHEMA` to run encoder-owned media and screenshot jobs in an optional dedicated Postgres schema, with automatic schema creation/migration when configured.
* **Bug Fixes**
  * Bundled self-host setup now better isolates encoder leadership to protect API uptime; schema-aware health draining now covers media and font transcode queues.
* **Documentation**
  * Updated install/architecture/media-optimizer docs and bundled Docker Compose with consistent `WPMGR_RIVER_MEDIA_SCHEMA` and upgrade/drain guidance.
* **Tests**
  * Added unit/integration coverage for schema qualification, query generation, config loading, and public-vs-media dual-schema isolation.


mosamlife (commented):
Thanks for this — it's a clean, well-documented fix for a genuinely nasty bug (the encoder winning River leader election and starving the API's `uptime_probe` / `cron_kick` periodics). We ran it through a correctness + security pass. **Verdict: approve with a few nits.** CI is green and the change is mergeable; we'd just like one item addressed before merge, plus a couple of optional follow-ups.

## What we liked
- **SQL identifier surface is fully closed.** Every schema/table/role value that reaches raw SQL (`CREATE SCHEMA` / `GRANT` / `ALTER DEFAULT PRIVILEGES`) is both regex-validated (`^[A-Za-z_][A-Za-z0-9_]*$`, rejecting dotted/quoted/dashed/leading-digit) **and** `pgx.Identifier{}.Sanitize()`-quoted in `riverutil/schema.go`. Queue names stay bound parameters. Nice defense-in-depth.
- **Least privilege is correct** — the app role gets only `USAGE` + table/sequence DML, no `CREATE`/ownership; `EnsureSchema` runs on the migration-owner pool, mirroring the existing `public` River grants (no privilege widening).
- **The second API client is provably inert** — no `Queues`/`Workers`/`PeriodicJobs`, never `.Start()`-ed, insert-only via `SkipUnknownJobCheck`. It can't win leadership in the media schema. The dual-schema integration test proving job routing (media/screenshot → media schema, health → public, inverse counts zero) is exactly the right thing to assert.
- Tenant isolation / RLS unaffected (River tables are infra, no tenant data crosses into the new schema). Boot ordering is correct and `EnsureSchema` is idempotent and fails loud.

## Please address before merge
**1. Compose default isolates on upgrade and can strand existing jobs.** `infra/docker-compose.yml` sets `WPMGR_RIVER_MEDIA_SCHEMA` to `media_encoder` by default for both services. That's great for fresh installs, but a self-hoster **upgrading** an existing deployment will have the encoder stop reading `public.river_job`, leaving any already-queued `media_encode` / `site_screenshot_capture` rows in `public` orphaned (no data loss, but they never run). Could you either:
   - add a short upgrade note to `.env.example` + `docs/install.md` ("on upgrade, drain media jobs first, or set this empty to keep current behavior"), or
   - default the compose value to empty (opt-in) so isolation is a deliberate switch?

   Either is fine — we just don't want a silent stranding on upgrade.

## Optional follow-ups (non-blocking, happy to take as a fast-follow)
2. **Validate the schema at config-load time.** An invalid `WPMGR_RIVER_MEDIA_SCHEMA` (e.g. a hyphen) currently fails at boot via `NormalizeSchema` rather than surfacing through `config.Validate`'s degraded-mode path, so it crash-loops instead of parking in readyz-degraded like the rest of the config. Threading `NormalizeSchema` into `config.Validate` would make the failure mode consistent.
3. **Log the effective media schema at startup on both binaries.** A mismatch between the API and the encoder silently strands media jobs. It's documented, but an INFO log of the resolved schema on each process makes an operator eyeball-diff trivial.
4. **Waker is missing `FontTranscodeQueue`.** The encoder's drain path counts three queues (`media_encode`, `site_screenshot_capture`, `font_transcode`), but the API waker (`cmd/wpmgr/main.go`) only passes the first two. This is latent today since `font_transcode` isn't fully wired through the API binary yet, but when it is, the waker won't hold the encoder awake for a font-transcode-only workload. Worth adding the queue now so it doesn't get missed.
5. **Test comment nit:** `TestRiverDualSchemaIsolation` is described as verifying the "leadership split," but it actually verifies job **routing** (the clients aren't started). Routing is the right thing to assert with insert-only clients — just reword the comment so it doesn't overclaim.

Once #1 is in we're good to merge. Thanks again — this is solid work. 🙏

BoostLabsAU:
Thanks for the thorough review! I've addressed the required item and folded in all four optional follow-ups.

**Required**
- **Upgrade stranding (#1):** Added explicit upgrade guidance in `.env.example` and `docs/install.md` (commit `6a343f8`). Existing deployments are told to drain media/screenshot jobs first, or keep `WPMGR_RIVER_MEDIA_SCHEMA` empty / `public` on both services until they are ready to switch. Kept the bundled compose default as `media_encoder` for fresh installs.

**Optional follow-ups (all in commit `5867f70`)**
- **Config-time validation (#2):** `WPMGR_RIVER_MEDIA_SCHEMA` is now validated in `config.Validate`, so an invalid identifier parks in readyz-degraded alongside the other config problems instead of crash-looping at River bootstrap. Added unit coverage for empty / `public` / valid / invalid values.
- **Startup schema log (#3):** Both the API and the media-encoder now log the resolved media schema at startup (`media_river_schema`), so a mismatch between the two processes is trivial to spot.
- **Waker font queue (#4):** The API waker now includes `font_transcode` alongside `media_encode` and `site_screenshot_capture`, matching the encoder drain queues so a font-transcode-only workload keeps the encoder awake.
- **Test comment (#5):** Reworded `TestRiverDualSchemaIsolation` so it describes job routing (insert-only clients, never started) rather than leader election.

**Verification**
- `go test ./internal/config ./internal/riverutil ./internal/media`, `go build ./cmd/wpmgr`, `go vet`, and `gofmt` all clean locally.
- The encoder image build and the container-based dual-schema integration test run in CI (CGO encoder + testcontainers), and the required checks are green.

No changelog churn beyond the existing entry, since these refine the same unreleased change. Happy to tweak any wording or split something out if you would prefer.

mosamlife (dismissed):

Verified the isolation is **correct and prod-safe**: the API's second River client is genuinely insert-only (no workers, no periodics, never `.Start()`ed), so it can't compete for leadership; the media-encoder leads its own schema. The schema name is regex-allowlisted (`^[A-Za-z_][A-Za-z0-9_]*$`) **before** any SQL interpolation and additionally pgx-identifier-quoted — I tried `public; DROP ...`, embedded quotes/NUL, leading digit, unicode; all rejected. GRANTs are minimal and scoped to the single media schema; privileged DDL uses the migration-owner pool, runtime inserts the unprivileged app pool. The empty-var path (GCP prod) is a byte-equivalent no-op, so this can't regress default single-schema deploys. No dependency changes, no new network/exec. Security: SAFE-TO-MERGE.

This genuinely fixes the "uptime goes silent when the encoder wins River leadership" bug. Thanks for the thorough tests (incl. the dual-schema routing integration test).

**Follow-ups (non-blocking):**
- **S1 — mismatch is a silent foot-gun.** If the API and encoder disagree on `WPMGR_RIVER_MEDIA_SCHEMA` (one set, one not), media/screenshot jobs enqueue into a schema nobody drains and stall silently. It's documented + logged at startup, which is reasonable, but a one-time self-check logging the row count in the "other" schema would harden it.
- **S2 — first-boot race.** API and encoder both run `EnsureSchema` (`CREATE SCHEMA IF NOT EXISTS`) at startup with no ordering; a truly-fresh DB can throw a transient catalog conflict and crash-loop one process once (self-heals on restart). A retry-on-duplicate, or letting only the API own `EnsureSchema`, would remove it.
- **N1 — release note:** this also quietly fixes a latent waker bug — the drain hold now counts `font_transcode` + `site_screenshot` queues, not just `media_encode`, so the encoder stays warm for screenshot/font work too. Worth a line in the release notes.

PR #95, opened 2026-06-22
fix(agent): apply page-cache variant settings

BoostLabsAU:
## What
- Align the control plane `perf_config_update` payload for the four page-cache variant lists with the agent's persisted and drop-in config keys: `bypass_urls`, `bypass_cookies`, `include_queries`, and `include_cookies`. Public API and database field names are unchanged.
- Add a pre-lookup bypass URL guard to the `advanced-cache.php` drop-in so configured bypass rules prevent serving an already-warmed cache file.
- Include `bypass_urls` in the rendered drop-in config via `CacheConfig::toDropinArray()`.
- Add regression tests for payload wire names, agent command persistence of unprefixed keys, and drop-in fast-path bypass behavior.
- Bump the drop-in version to `0.45.1` so existing installs reinstall the updated drop-in on the next config push.
- Add a `CHANGELOG.md` `### Fixed` entry under `## [Unreleased]`.

## Why
The control plane stores advanced page-cache rules as `cache_bypass_urls`, `cache_bypass_cookies`, `cache_include_queries`, and `cache_include_cookies`, but the agent command whitelist and `CacheConfig` only recognized the unprefixed names. Saves appeared successful but the lists were silently ignored. The drop-in also did not enforce bypass URLs before serving an existing cache file, so a bypass rule could still return a cache HIT for a previously cached page.

**Risk:** This touches the page-cache serving fast path. The bypass URL check uses the same case-insensitive substring matching as the PHP `Cacheability` layer, applied to the request path, and runs after the existing method/cookie guards.

## Checklist
- [ ] `make lint` passes
- [ ] `make test` passes
- [ ] `make build` passes
- [x] OpenAPI updated (if endpoints changed) - no endpoint changes
- [x] Docs updated (if behavior changed) - changelog entry added; no other docs needed for this fix
- [x] Touches auth / crypto / agent protocol / RBAC / tenant isolation? Yes - agent protocol and the cache serving fast path. Flagging for security review; please take a look at the cache-bypass behavior.

## Testing
I ran the focused Go and agent suites plus PHPCS against the changed paths. I don't have the Docker-based `make agent-plugincheck` set up locally, so I'm relying on CI for the full lint/test/build and plugincheck gates.

- `go test ./internal/perf` (from `apps/api`): passes, including a new regression that fails if the cache variant lists ever marshal back to the `cache_*` names.
- Agent PHPUnit `CacheCommandsTest`, `CacheConfigAndRolesTest`, `CacheabilityTest`, and `AdvancedCacheDropinTest` (from `apps/agent`): 48 tests, 149 assertions, all passing.
- PHPCS clean on the changed agent files.
- `node apps/landing/scripts/check-copy.mjs --extra CHANGELOG.md` and `git diff --check`: clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed advanced page-cache bypass and cache variant settings not being properly delivered to the caching system. Configured URL bypass rules (including case-insensitive matching), cookie settings, and query include/exclude behavior now correctly influence which cached drop-ins are served.

* **Tests**
  * Added/expanded test coverage for bypass URL behavior, persistence of unprefixed variant keys, ignoring of prefixed aliases, and correct request payload key names sent from the performance config pipeline.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ House rules: no em dashes, no en dashes, no competitor names. Use "to" for range
 ### Fixed
 
 - Backups now preserve plugin and theme vendor code in directories named `cache`, `upgrade`, or `upgrade-temp-backup` while still excluding runtime cache and update staging roots.
+- Advanced page-cache bypass and variant settings now reach the WordPress agent so saved URL, cookie, and query rules affect the rendered cache drop-in.
 
 ## [0.57.0] - 2026-06-21
 
--- a/apps/agent/assets/wpmgr-advanced-cache.php
+++ b/apps/agent/assets/wpmgr-advanced-cache.php
@@ -34,7 +34,7 @@
  * finds in the on-disk copy; a mismatch triggers a transparent reinstall so
  * existing sites always run the current drop-in logic without manual intervention.
  */
-define('WPMGR_PAGE_CACHE_DROPIN_VERSION', '0.45.0');
+define('WPMGR_PAGE_CACHE_DROPIN_VERSION', '0.45.1');
 
 if (!defined('ABSPATH')) {
     exit; // No direct access.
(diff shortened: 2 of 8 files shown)


mosamlife (approved):

Verified the fix is in the **right direction**: the agent's `perf_config_update` whitelist (`KNOWN_KEYS`) and `CacheConfig` read the unprefixed keys (`bypass_urls`, etc.), so the CP sending `cache_bypass_urls` was silently dropped — saves succeeded, lists never applied. Renaming only the CP→agent wire JSON tags is correct; public API, DB columns, and OpenAPI keep the prefixed names. The new drop-in bypass check runs after `$wpmgr_uri` is sanitized and only ever `return false` (bypass → boot WordPress), so it can reduce cache hits but never poison/serve a wrong page. Tests are substantive (they assert the unprefixed wire names and reject the prefixed aliases). Security: no supply-chain/injection/scope-creep — SAFE-TO-MERGE.

**Optional follow-up (non-blocking nit):** the comment says the fast-path match is "identical to the PHP cacheability path," but the fast path matches the **query-stripped** path while the write layer (`Cacheability::matchesBypassUrl`) matches the **full URI including the query string**. Behavior is still safe (a query-only bypass keyword just degrades to a normal miss instead of short-circuiting), but the comment slightly overstates the equivalence — worth a one-line tweak, or match the query-stripped path in `Cacheability` too for true parity. Not a merge gate.

PR #106, opened 2026-06-23
fix(agent): apply CDN rewrite settings

BoostLabsAU:
## What

Fixes propagation of saved CDN rewrite settings so the WordPress agent rewrites static asset URLs with the operator-selected scope. No public API, UI, or database field names change.

Changes:
- `apps/api/internal/agentcmd/cache_contract.go`: the `perf_config_update` payload now sends the CDN enabled flag under the agent-side key `cdn` instead of the control-plane key `cdn_enabled`. `cdn_url` and `cdn_file_types` remain unchanged.
- `apps/api/internal/perf/service.go`: adds `mapCDNFileTypes()` to translate operator-facing values to agent enum values before building the push payload:
  - `images` to `image`
  - `css_js` to `css_js_font`
  - `all` to `all`
  - empty or unknown values to `all`
- `apps/api/internal/perf/beacon_push_test.go`: adds regression coverage for the wire key mapping, file-type enum mapping, CDN URL inclusion, and absence of provider or credential fields in the JSON payload.
- `apps/agent/tests/CacheCommandsTest.php`: adds regression coverage proving `perf_config_update` persists `cdn`, `cdn_url`, and mapped `cdn_file_types` into `wpmgr_perf_config`.
- `apps/agent/tests/OptimizerTransformsTest.php`: adds regression coverage proving `css_js_font` rewrites CSS, JavaScript, and font URLs without rewriting image URLs.
- `CHANGELOG.md`: adds one `### Fixed` entry under `## [Unreleased]`.

## Why

The control plane stored and sent `cdn_enabled`, while the agent persisted and read `cdn`. `PerfConfigUpdateCommand::maybePersistOptimizationConfig()` intersects the incoming payload with `PerfConfig::toArray()`, so `cdn_enabled` was dropped and the optimizer saw CDN disabled. With a working CDN URL, a fresh render still produced origin asset URLs. This change aligns the wire contract and maps the file-type enum so saved CDN settings reach the agent.

Credentials remain control-plane-only. The agent payload does not include provider, token, zone ID, or encrypted credential fields.

## Checklist

- [ ] `make lint` passes
- [ ] `make test` passes
- [ ] `make build` passes
- [x] `go test ./internal/perf -count=1` passes (from `apps/api`)
- [x] `composer test -- tests/CacheCommandsTest.php tests/OptimizerTransformsTest.php` passes (from `apps/agent`)
- [x] `composer lint -- tests/CacheCommandsTest.php tests/OptimizerTransformsTest.php` passes (from `apps/agent`)
- [x] `node apps/landing/scripts/check-copy.mjs --extra CHANGELOG.md; git diff --check` passes
- [ ] OpenAPI updated (if endpoints changed)
- [ ] Docs updated (if behavior changed)
- [x] Touches auth / crypto / agent protocol / RBAC / tenant isolation? Flagged for security review because this touches agent protocol.

## Summary by CodeRabbit

* **Bug Fixes**
  * CDN rewrite settings now correctly transmit to the WordPress agent with the selected asset type scope. CDN credentials remain control-plane only.

* **Tests**
  * Added test coverage for CDN configuration persistence and asset-type-specific rewriting. Tests verify correct behavior when CDN file types are configured for CSS, JavaScript, fonts, and all asset types.


mosamlife (approved):

Verified the fix is the right direction and complete:
- The agent reads/persists the **unprefixed** `cdn` / `cdn_url` / `cdn_file_types` (`class-perf-config.php`), so the CP sending `cdn_enabled` was silently dropped by the `intersect(toArray())` filter — exactly the mechanism behind the silent-ignore class (same as #95). Renaming the wire tag to `cdn` is correct.
- `mapCDNFileTypes` maps the operator enum (`images`/`css_js`/`all`) to the agent's accepted set, which is exactly `['all', 'css_js_font', 'image']` (`class-perf-config.php:469`, consumed as a fixed `GROUPS` regex-map key, never interpolated). Before this, `images`/`css_js` were rejected by the agent and fell back to `all`, so the operator's scope was ignored. The `default: "all"` is a safe fallback.

Security: **SAFE-TO-MERGE**. The push carries only the non-secret rewrite config; no provider/token/zone/credential field is added (the `CDNCredentials` struct stays CP-only, used solely on the purge path). The new `beacon_push_test.go` deliberately sets a provider on the input and asserts it never reaches the wire JSON — a real leak-regression guard. The file-type value is allowlisted on both the CP (validated on save) and the agent (re-validated), and only ever selects a map key, so nothing attacker-influenced reaches a regex or URL build. Signed-command transport unchanged.

Thanks for the precise root-cause and the negative-assertion test.

PR #121, opened 2026-06-28
fix(api): ~8s post-idle /sites = dead pgx connection hang (validate-on-acquire + keepalive + phase timing)

mosamlife:
The 8s was confirmed INSIDE the Go process (app middleware timer = 8.15s, concurrent /sites = 53ms, same warm instance). pgx handed out a stale connection (VPC-dropped) and the query hung on the dead socket ~8s. Fix: BeforeAcquire ping-validates every conn (500ms) so a dead one fails fast and pgx dials fresh; TCP keepalive; shorter idle eviction. Plus phase-timing logs to prove it. Already deployed as api 0.61.6.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added automated daily cleanup of old uptime-probe data (90-day retention) to control database growth.
* **Performance Improvements**
  * Improved fleet uptime query speed with a new covering database index and a short in-process cache for repeated requests.
* **Bug Fixes**
  * Enhanced database connection resilience against stale/half-open sessions to reduce intermittent request failures.
* **Monitoring & Diagnostics**
  * Added structured logging for site listing, including warnings for slow requests with key timing breakdowns.

### DIFF
--- a/apps/api/cmd/wpmgr/main.go
+++ b/apps/api/cmd/wpmgr/main.go
@@ -410,6 +410,7 @@ func run(ctx context.Context, cfg config.Config, logger *slog.Logger) error {
 
 	tenantSvc := tenant.NewService(tenant.NewRepo(pool), validator, clock)
 	siteSvc := site.NewService(site.NewRepo(pool), validator, clock)
+	siteSvc.SetLogger(logger)
 	auditRec := audit.NewRecorder(pool, clock)
 
 	// A narrow tenant-creation capability handed to the auth domain (bootstrap +
--- a/apps/api/cmd/wpmgr/main.go
+++ b/apps/api/cmd/wpmgr/main.go
@@ -494,6 +494,12 @@ func run(ctx context.Context, cfg config.Config, logger *slog.Logger) error {
 	// is not configured — in that case only DB rows are pruned.
 	fileTransfersGCWorker := files.NewFileTransfersGCWorker(pool, nil, logger) // deleter wired below if S3 enabled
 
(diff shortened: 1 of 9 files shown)


mosamlife:
Reviewed the CodeRabbit feedback:

**1. `gc_worker.go` retention delete won't use an index (Major) — correct, fixed.** Confirmed: after m85 the `site_uptime_probes` indexes lead on `site_id` / `tenant_id` / `(site_id, tenant_id)`; none lead on `probed_at`, so `DELETE ... WHERE probed_at < cutoff` would sequentially scan the table. On a large table that daily prune would also evict the buffer-cache pages the fleet-uptime query depends on — re-introducing the cold-cache latency m85 just removed, once per day. Fixed in **m86** (`20260719000000_m86_uptime_probes_retention_idx.sql`): a btree index on `(probed_at)` so the prune does an index range scan over just the old rows. Shipped as api 0.61.8.

**2. `db_test.go` reimplements the prod config (Trivial nitpick) — valid, deferring.** Fair point that the test duplicates the `connect()` tuning rather than asserting the production path, so it wouldn't catch the tuning being removed. Low priority; noted as a follow-up rather than blocking.

**3. ast-grep "SQL injection" on the `fmt.Sprintf` DELETE — false positive.** The only interpolated value is `probeGCBatchSize` (`const = 500_000`, a compile-time int); the actual filter value `cutoff` is parameterized (`$1`). No user input reaches the query string, so there's no injection. (Kept the `fmt.Sprintf` since `LIMIT` can't be a bind parameter in this position.)

PR #355, opened 2026-08-06
feat(wporg): prepare the WordPress.org listing, assets and publish workflow

mosamlife:
Fleet Agent Site Manager was **approved on WordPress.org today**. Trunk is empty and nothing has been published.

This prepares everything that must be right *before* the first commit, because a wp.org publish cannot be undone: a version number can never be reused, and a wrong `Stable tag` ships to every installed site.

**Nothing is published by this PR. No `svn` command was run.**

## The readme, rewritten for a stranger

The old one was written to pass review, and it did. It opens with signature verification and command allow-lists, which is the wrong register for a page where a visitor decides in ten seconds.

The rewrite follows what actually works, checked against the real analogues rather than assumed:

- **MainWP Child**, the closest thing that exists to this product, puts the dependency **in the title**
- **Plausible** states "you need our service" in one bolded sentence
- **ManageWP Worker** handles its name mismatch in the FAQ, not the description

An earlier draft explained the three names (WPMgr / WPMgr Agent / Fleet Agent Site Manager) at length and was refused for it: **explaining a name is what a rebrand does.** The title now carries the dependency, the requirement is one sentence, and the naming question is FAQ #2.

Every disclosure the reviewers required is intact: Privacy, External services with all twelve services and their Terms and Privacy URLs, the inert-until-configured statement, and the credits. **The register changed, not the disclosures.**

No claim the shipped build cannot support: the wp.org build physically excludes the self-updater, so nothing describes agent self-update, and vulnerability matching and uptime probing are attributed to the dashboard because that is where they run.

**Installation was wrong** and is fixed. The real flow is Save URL, then Pairing code, then Enroll. The old steps named a token and a Connect button that do not exist. Upgrade Notice was seventy releases stale.

## The publish workflow: manual, and gated

Not automatic, and the repo's own history is the argument: **eight of the last nine version tags did not move the agent at all**, so an automatic publish would have tried to create `tags/0.61.119` eight times. Automation whose main job is deciding *not* to run, on a path where being wrong is permanent, is the wrong trade.

`workflow_dispatch`, two jobs. `build` packages, runs the real Plugin Check, and refuses if the tag already exists upstream. `publish` carries `environment: wordpress-org`, so it stops and waits for human approval **after** Plugin Check is green.

Credentials are **environment**-scoped, not repository-scoped, and that is the point: only a job declaring the environment can read them, and declaring it is exactly what forces the approval. The credential and the gate cannot be separated. The `build` job runs with no credentials at all.

## A defect that would have fired on launch day

The workflow probed for `svn --password-from-stdin` using `svn help commit`. That **hides global options**, so the probe answers NO on every Subversion that has the flag. Measured on 1.14.5:

```
svn help commit    | grep password-from-stdin   ->  no match
svn help commit -v | grep password-from-stdin   ->  match
```

The step could only fail *after* Plugin Check went green and *after* the approval was clicked, while blaming the runner for a problem it did not have.

## Stale versions, swept

`README.md` said **v0.61.94**, thirty-three releases behind, including three `docker pull` commands people copy and run. The marketing hero badge on wpmgr.app said **v0.61.112**. Both are linked from the listing. Both now 0.61.127, matching the agent.

Roughly a dozen other references are **historical provenance** (`"Reported against 0.61.102"`, `"Absent from every agent before 0.61.110"`) and were deliberately left, because rewriting them would make them false.

## The copy gate was never running in CI

`check-copy.mjs` only ever ran inside `pnpm -C apps/marketing build`, which `ci.yml` never invokes. **The rule was enforced nowhere.** It now runs in CI and covers 366 files: the marketing site, `apps/agent/readme.txt` which *is* the public listing page, the plugin header, and the agent PHP carrying user-facing strings.

It immediately caught three things already shipping: an em dash in a translatable 2FA backup-codes string, an em dash in the admin screen, and a competitor product name in a comment inside the wp.org zip.

`README.md` carried **195 em dashes and 6 en dashes**; all gone, with ranges now reading "to" per the CHANGELOG's own house rule. 186 lines changed in place, no additions or deletions, all structural markers balanced.

## Assets

Icons and banners are generated by tracked scripts from the existing brand files, at the exact wp.org filenames and dimensions:

```
icon-128x128.png   128x128      banner-772x250.png    772x250
icon-256x256.png   256x256      banner-1544x500.png   1544x500
```

**Nothing was designed.** `apps/marketing/public/icon-512.png` was already the right composition and the mark already avoids any WordPress letterform, which is the trademark constraint. Output lands in the gitignored `release/wporg-assets/` so it is reproducible without living in git.

**Screenshots are the one thing only a human can produce.** The shot list is in `release/wporg-assets/README.md`, and the readme is the single source of truth for their order after a conflict where two lanes wrote different lists and each deferred to the other.

## Gates

```
agent phpunit  3176 pass    phpcs 0    Plugin Check 0 errors
go build / ./internal/...   clean
web tsc 0    marketing typecheck clean
check-copy   passes across 366 files
workflow YAML validates
```

## Summary by CodeRabbit

- **Documentation**
  - Updated plugin branding, version information, installation and pairing guidance, requirements, integrations, privacy details, FAQs, screenshots, and changelog through version 0.61.127.
  - Added a WordPress.org publishing and release operations runbook.

- **Bug Fixes**
  - Replaced em-dash placeholders in the admin status panel with “Not set” when Site ID or Tenant ID values are missing.
  - Improved backup-code warning text for readability.

- **Chores**
  - Updated marketing copy and displayed product version to v0.61.127.
  - Added validation for WordPress.org listing content and publishing assets.


mosamlife:
Pushed `8731639` addressing all four review findings. Each turned out to be the same shape, which is worth naming: **something reporting success without having checked**, on paths where the mistake is public and hard to walk back.

**The copy gate was scanning one byte.** `indexOf("*/")` returns -1 on an unterminated docblock, and `slice(0, -1 + 2)` is the first character of the file. The loop then matched nothing, found no violations, and still counted the header as scanned, so the summary printed "1 plugin header" having read one byte. Measured before fixing:

```
Plugin Name: Streamline Fleet
Description: A world-class seamless thing
-> check-copy passed
```

Now fails closed unless the slice actually contains a `Plugin Name:` field, which also catches a `*/` that closes some unrelated block first.

**The runbook contradicted the workflow's security model.** It said repository secret; the workflow is built on `environment: wordpress-org` precisely so the credential and the approval cannot be separated. A repository secret is readable by every workflow including ungated ones, so following the runbook would have quietly dissolved the property the workflow exists to provide.

The live configuration was already correct, verified rather than assumed:

```
repos/.../actions/secrets                        -> (empty)
repos/.../environments/wordpress-org/secrets     -> WPORG_SVN_PASSWORD
repos/.../environments/wordpress-org/variables   -> WPORG_SVN_USERNAME
protection_rules                                 -> required_reviewers
```

So this was a document that would have misled the next person, not a live hole. Sections are also reordered, since the old 0.2 and 0.3 asked you to create secrets inside an environment that 0.4 had not created yet. Added the two `gh api` calls that check scope, because both scopes look identical from the workflow's side right up until the day it matters.

**`verify.py` now runs before publish.** Dimensions, file size and opacity each fail the same silent way on wp.org: the commit is accepted and the asset is omitted or mangled. The step checked only names and extensions.

That also closes a destructive case I had not accounted for. Publish mirrors with `rsync --delete`, so staging a **partial** set does not add to the published assets, it **deletes** the icon and banner. Absent is safe and leaves the listing untouched; half-present is not. `verify.py` requiring all four is exactly the right guard. Separately, `mode=assets-only` with no assets directory now fails rather than "succeeding" having done nothing.

**A latent bug found while testing that.** The uppercase-filename check is locale dependent. Under a UTF-8 collation locale the range `[A-Z]` follows dictionary order (aAbBcC...), so it matches lowercase too:

```
$ LANG=en_IN.UTF-8   find pa -name '*[A-Z]*'   ->  pa/screenshot-1.png
$ LC_ALL=C           find pa -name '*[A-Z]*'   ->  (nothing)
$ LC_ALL=C           find ok -name '*[A-Z]*'   ->  ok/Icon-Bad.png
```

Runners default to `C.UTF-8` where it behaves, so it would have sat dormant until a runner image changed and then failed a publish while naming a file that was perfectly correct. Pinned to `LC_ALL=C`.

Every path was exercised rather than read: unterminated header fails closed, valid header still passes 366 files, partial assets fail on MISSING, full set passes, genuine uppercase still caught, `assets-only` with no directory exits 1, `release` with no directory warns and exits 0, workflow YAML parses with `publish` gated and `build` holding no credentials.

PR #370, opened 2026-08-06
feat(seo): send real article metadata, and stop faking dateModified

mosamlife:
Tier 2 item **2.7**. Stacked on #368 (branched from it while Actions was down); retarget to `main` once that merges.

Every blog post and guide was published as a generic web page. That is a prerequisite problem rather than a cosmetic one.

## What was wrong

`lib/seo.ts` hardcoded `type: "website"` for all callers, so **all 12 articles shipped `og:type=website`** with no `article:published_time` and no `article:modified_time`. Nothing downstream could distinguish an article from a landing page.

## The `dateModified` default was worse than missing

`buildArticleLd` did `dateModified: dateModified ?? datePublished`, and **no caller ever passed one**. So every article asserted it was last modified on the day it was written.

That is not a harmless default. It is a claim, it becomes false the first time a post is edited, and it makes a genuine refresh **indistinguishable from no refresh**. Refreshing underperforming posts is the entire weekly half of the planned automation loop, and it cannot signal anything through a field that already claims to be current.

Both `dateModified` and `article:modified_time` are now emitted **only** when a piece carries an explicit `updated` frontmatter date. Absent means never revised: true and useful, rather than false and useless.

`updated?` is optional on both frontmatter types, so no existing content file needs editing and nobody is tempted to invent a date to satisfy a required field.

## Found while here

Both JSON-LD `Article` builders were still passing a **trailing-slash url** (`/blog/${category}/${slug}/`). Missed by d5d2a57 and a99c6eb because those swept canonicals, the sitemap and `href` attributes but not string arguments to the LD builders. The structured data was naming a URL that 308-redirects. Fixed in both.

## Verified in the built HTML

Default state:

```
og:type                  article
article:published_time   2026-06-15
article:section          Security
article:tag              security, login, 2fa
article:modified_time    absent      (nothing revised yet)
LD url                   https://wpmgr.app/blog/wordpress-security/harden-wordpress-login
LD dateModified          absent
```

Then, because "it will work when set" is not a verified claim, I set `updated: "2026-08-06"` on one post and rebuilt:

```
article:modified_time    2026-08-06
LD dateModified          2026-08-06
```

Probe reverted, tree rebuilt clean. typecheck / lint / build / check-copy all pass.

### DIFF
--- a/apps/marketing/app/blog/[category]/[slug]/page.tsx
+++ b/apps/marketing/app/blog/[category]/[slug]/page.tsx
@@ -35,6 +35,13 @@ export async function generateMetadata({ params }: Props): Promise<Metadata> {
     description: fm.description,
     canonical: `/blog/${category}/${slug}`,
     ogImage: ogImageUrl,
+    article: {
+      published: fm.date,
+      ...(fm.updated ? { modified: fm.updated } : {}),
+      authors: [fm.author ?? "WPMgr Team"],
+      section: CATEGORY_LABELS[category] ?? category,
+      ...(fm.tags ? { tags: fm.tags } : {}),
+    },
   });
 }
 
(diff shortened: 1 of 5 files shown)


mosamlife:
Superseded by #371, which contains this commit (`aa3091b`) as an ancestor. Closed automatically when its base branch was deleted during the stack merge, and GitHub will not reopen a PR whose base is gone. The work is not lost: verified with `git merge-base --is-ancestor aa3091b origin/feat/internal-linking-and-blog-cta`.

PR #392, opened 2026-08-09
fix:update SmtpHandlerTest

### LINKED ISSUE #380: SMTP fails on some website after successful configuration days ago
**Describe the bug**: smtp continue fails with no reason on website that already have configured all the settings and have already sent successful email, test email or normal email (form, wordpress udpate etc).

**To reproduce**: only certains website fail the sending "wordpress udpate to" email, but also a test of smtp fails. In some website changing the password and reinsert it into the settings make the smtp work again, in others no. So it seems to be a bug agent related.

**Environment**: 0.61.127, self hosted

Image
Image

iacopoincerpi:
## What

Fixes #380

Resolved SMTP authentication failures during automated background tasks by ensuring credentials/connection state are refreshed before dispatch.

## Why

Emails triggered automatically by WordPress background tasks (e.g., core updates like "Your site has been updated") fail with `SMTP Error: Could not authenticate` after prolonged periods or dormant connection intervals (e.g., 21-day minor release cycles).

## Checklist

- [x] `make lint` passes
- [x] `make test` passes
- [ ] `make build` passes
- [ ] OpenAPI updated (if endpoints changed)
- [ ] Docs updated (if behavior changed)
- [ ] Touches auth / crypto / agent protocol / RBAC / tenant isolation? → flag for security review

## Summary by CodeRabbit

* **New Features**
  * Added SMTP email delivery with support for authentication, encryption, custom sender and recipient details, message bodies, headers, and attachments.
  * Invalid recipients are skipped while valid recipients continue receiving messages.
* **Bug Fixes**
  * SMTP requests now report structured success and failure results.
  * Delivery fails clearly when no valid recipients remain.

### DIFF
--- a/apps/agent/tests/Email/SmtpHandlerTest.php
+++ b/apps/agent/tests/Email/SmtpHandlerTest.php
@@ -13,207 +13,213 @@
 
 declare(strict_types=1);
 
-namespace WPMgr\Agent\Tests\Email;
+namespace WPMgr\Agent\Email\Handlers;
 
-use PHPMailer\PHPMailer\FakePhpMailerRegistry;
-use PHPUnit\Framework\TestCase;
-use WPMgr\Agent\Email\Handlers\SmtpHandler;
-
-require_once __DIR__ . '/fake-phpmailer.php';
+use WPMgr\Agent\Email\AddressParser;
+use WPMgr\Agent\Email\ProviderHandlerInterface;
 
 /**
- * @covers \WPMgr\Agent\Email\Handlers\SmtpHandler
+ * SMTP provider handler using WP's bundled PHPMailer.
  */
-class SmtpHandlerTest extends TestCase
-{
-    private function base_mail(): array
-    {
-        return [
-            'to'          => ['recipient@example.com'],
-            'cc'          => [],
-            'bcc'         => [],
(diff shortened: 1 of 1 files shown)


mosamlife:
Thank you for this, and for the original report in #380. Both were genuinely useful, and the report is what made the fix findable.

Closing this one because the bug turned out to sit a layer below the test. The change here updates `SmtpHandlerTest`, but the handler was only the messenger: by the time it authenticated, the password had already been deleted from the site. The real fault was in the wire contract between the control plane and the agent. The control plane distinguishes "the operator supplied no secret" from "clear the secret", and its SQL preserves the stored credential in the first case, but the field it sends the agent could not carry that distinction, so the agent received the same bytes for both and deleted a working password on an ordinary settings push.

The fix went out in 0.61.131 and covers the whole path, including the case your test was pointing at. It also stops the agent acting on a settings update it cannot fully parse, and makes it report a settings write that did not land rather than returning success.

Your report contained the detail that cracked it: that re-entering the same password fixed some sites and not others. That is not what a wrong password looks like, and it is what separated a credential being deleted from a credential being wrong. Sites recover on their own once the plugin updates.

If you would like to send a follow-up PR at some point, please do.

PR #401, opened 2026-08-10
fix(api): clear the confirmed bot review debt on the social sign-in path

mosamlife:
Bot-review debt from the social sign-in PRs, which were merged with `--admin` before their reviews finished. Rebased onto today's `main` and re-verified rather than trusted, since main has moved 29 commits since these were written.

Every item below is a confirmed finding, not a style note. The theme is that the audit log and the stored identity record were both quietly lying.

## What was wrong

**Credential changes went unrecorded for the accounts with the least oversight.** Setting a first password, or disconnecting a provider, was skipped entirely when the account had no organisation. That is not an edge case: it describes a site collaborator, a portal user, a brand new account created by signing in with a provider, and anyone whose only organisation is inside its deletion grace window.

**Connecting a provider was indistinguishable from using one.** Connecting a provider to an account, creating an account from one, and a stored record changing issuer were all filed as "Signed in with SSO", so a credential change looked like an ordinary login to anyone scanning or filtering the log.

**A provider that stops reporting an address erased the one it had reported.** GitHub returns nothing when a user makes their address private, and that empty report was written straight over the stored value, so the sign-in meant to keep the record current destroyed it. The `TouchIdentityLogin` query now leaves both columns alone on an empty report and stamps only the login, which is the one fact that sign-in actually established. `email_verified` moves with the address it describes and never on its own, because a stored address carrying a flipped verified flag is a worse record than either value alone.

**An issuer differing only in capitalisation or a trailing slash updated nothing.** The sign-in worked, but the last-used stamp and the current address silently did not move, so a connected account could show as never used no matter how often it was used.

**A refused connect attempt on an unverified account destroyed the plan it signed up for.** The verification mail it sends replaces the previous one, so sending it without the plan did not merely omit the plan, it discarded it, and the account landed on free after verifying.

Also: Google sign-in is now bounded by a timeout as GitHub already was, connecting a second account from an already-connected provider reports a conflict rather than a 500, a provider switched off mid-flow no longer loses the page you were heading to, and the provider list is always a list rather than `null` on an instance with none configured, which is what the published API says.

## On the generated file

`internal/db/sqlc/user_identities.sql.go` is a real regeneration, not a hand-edit. The `.sql` source changed alongside it, and the embedded SQL is a faithful expansion of it: `sqlc.arg(email)` becomes `$1` and is correctly reused in both branches, the remaining parameters renumber accordingly, and the params struct reorders to match.

## Verified

Rebased cleanly apart from `CHANGELOG.md`, resolved by splicing these entries into the current `[Unreleased]` rather than overwriting today's release section. While there I corrected a stale count in an unrelated entry: the guard suite is 76 cases, not 75.

`gofmt` clean on every changed file, `go build`, `go vet` clean. CI runs the full suite.

## Summary by CodeRabbit

- **New Features**
  - Added clearer audit labels for social sign-ins, account linking, registrations, and identity changes.
  - Social provider responses now consistently return an empty provider list when none are configured.

- **Bug Fixes**
  - Preserved identity email details when providers omit an email address.
  - Improved duplicate-account conflict handling and audit routing.
  - Preserved valid return links after disabled-provider sign-in failures.
  - Retained selected plans during verification retries.
  - Added time limits to Google sign-in requests.
  - Strengthened public URL validation for safer redirects.


mosamlife:
The three open review threads on `validate.go` all report the same thing, and it was real: the public base URL format check sat behind `cfg.Social.Configured()`, which is off on essentially every install, so the value was examined only where it was least likely to be wrong.

It is fixed on the current head (`69c98b0`). The threads point at lines 307, 327 and 329, which are from the commit before the fix; the check now lives at line 203 inside `Advisories`, out of `validateSocialConfig` entirely, because leaving a non-social check inside a function named for social config is how the next person concludes it only matters to social.

Two details worth recording, because both nearly went the wrong way.

**The first attempt did nothing, and my own test hid it.** I edited the block reached only from `Advisories` and asserted against `Validate`. The test written to prove the fix passed against code with no effect. It only surfaced because the test was run rather than assumed. Both new tests now fail when the gate is put back:

```
--- FAIL: TestPublicBaseURLIsJudgedWithoutSocialSignIn
    "https://manage.example.com/?next=x" passed while social sign-in was off
    "https://manage.example.com/#frag"   passed while social sign-in was off
    "manage.example.com"                 passed while social sign-in was off
```

**An empty value is deliberately still a social-only complaint.** `TestAdvisoriesIgnorePublicBaseURLWithoutSocialSignIn` already pinned that, and it is the right line: nothing outside social requires this to be set, so ungating wholesale would newly fail every self-hosted install that never set it. A guard that cries wolf gets switched off, and then it guards nothing. Empty stays social-only; a value that IS set is judged for everyone.

PR #421, opened 2026-08-14
chore(harness): make the guards fail loudly instead of quietly

mosamlife:
An audit of the guards found this repository's signature defect — a check that announces success over
its own errors — in eleven more places after the `rm -rf` incident was closed. Each was reproduced with
an executed command before it was touched, and each fix was then attacked.

## What was silently permitted before this

```
sed -i 's/a/b/' apps/api/migrations/*.sql              PERMIT, 0 bytes   (all committed migrations)
cd apps/api && sed -i … migrations/<applied>.sql       PERMIT, 0 bytes
apps/api/../api/migrations/<applied>.sql               PERMIT, 0 bytes
cd apps/api/internal/db/sqlc && echo x > db.go         PERMIT, 0 bytes
python3 -c "shutil.copy('/tmp/x', '…/sqlc/db.go')"     PERMIT, 0 bytes
node -e "fs.openSync(p,'w')"                           PERMIT, 0 bytes
route-guard with no cwd in the payload                 PERMIT everything, 0 bytes
jq absent                                              BOTH guards silent on every hard deny
```

Seven of those were proved destructive by execution — target file checksum before, run, checksum after.
The interpreter arm was additionally polarity-inverted: it denied *reading* the generated sqlc tree and
permitted *overwriting* it, which is the hand-sync that caused a production 500 here.

## What was reporting success over failure

`route-guard-coverage.sh` scored a completely broken guard at `ask 0 (0.0%) / deny 0 (0.0%) / pass 248
(100.0%)` and exited 0, in both arms — a number that gets pasted into `docs/harness.md` as evidence.
`commit-gate.sh` exited 0 in silence when the agent's only uncommitted file had a name git quotes.
`harness-reap.sh` read a failed `du` as "under the ceiling, leaving it". Two `--help` outputs sliced a
hardcoded line range and dropped real content from the README's install script.

## And assertions that tested nothing

Four assertions in the `route-guard state` block survived deletion of the code they name; two of them
authorise deletes, and an auditor destroyed real file content with a mutated guard while the suite
stayed green. The suite's only two Python assertions both contained the literal `open(`, so a bare
substring match on `open(` would have passed them while shipping the inverted arm.

That defect recurred *inside this work* twice, and both times mutation testing found it rather than
review: one round shipped four new assertions that pinned nothing, the next shipped two. Both are fixed
and both are in the mutation table.

## Approach

`cd` tracking is now one shared `cd_track`, used by the write arms and the push arm alike, rather than
the push arm's private copy plus nothing elsewhere. The suite asserts the same `cd` shapes against both,
so they cannot drift apart silently — a mutation making the tracker guess reddens a push assertion and a
write assertion in the same run.

Where the effective directory genuinely cannot be determined, the last arm **asks**. It does not deny,
because after an unresolvable `cd` every relative path could be inside a protected tree and denying
reddens `cd "$(mktemp -d)" && echo x > out.txt`. It does not stay silent either — silence is the one
outcome removed.

## Proof

```
$ bash scripts/claude/guards_test.sh
guards_test: 950 assertions passed          (main: 419)
```

Every fix was mutated — deleted or inverted — and the suite confirmed to redden naming the right
assertion; the mutation tables are in the commits. The push arm was regression-checked by replaying
every payload the suite covers against the guard before and after:

```
total payloads replayed: 466
push-shaped payloads:    115
(no push decision changed on any shape the suite covers)
```

## Known gaps, stated rather than implied

Which migrations are *applied* is read from `HEAD` starting at the payload's cwd, not the effective
directory, so `cwd=/tmp` plus `cd <repo>/apps/api && sed -i migrations/<applied>.sql` still passes. The
sqlc, generated-tree and dead-app arms are fixed for that case; the applied-migration arm is not.

`bash-guard.sh` matches command text and cannot decide `eval`, `bash -c`, or a path built at run time.
It says so in its own comments now instead of implying coverage it lacks. `.githooks/pre-push` is the
lock; this is the early stop in front of it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Security guards now fail closed when required tooling is unavailable, preventing unapproved operations.
  * Improved handling of relative paths, directory changes, redirections, migrations, and branch-aware push checks.
  * Commit checks now accurately handle untracked, renamed, and copied files.
  * Cleanup and cache measurement failures now report errors without masking command results.

* **Improvements**
  * Help output is more complete and consistently generated across setup and maintenance scripts.
  * Vocabulary parsing now rejects missing, duplicate, or multi-line banned assignments.
  * Routing and coverage checks now reject invalid or unknown decisions.

### DIFF
--- a/.claude/agents/docs-writer.md
+++ b/.claude/agents/docs-writer.md
@@ -24,8 +24,10 @@ command is missing, say so and stop, never report a step you did not run.
 banned_counts() {
   f=${1:-.github/workflows/ci.yml}
   [ -f "$f" ] || { echo "cannot read $f: the vocabulary gate moved, so find it before trusting any count" >&2; return 1; }
-  b=$(grep -oE "banned='[^']+'" "$f" | sed "s/banned='//; s/'$//")
-  [ -n "$b" ] || { echo "no single-line banned='...' in $f: renamed, requoted or wrapped, which is not an empty list" >&2; return 1; }
+  raw=$(grep -oE "banned='[^']+'" "$f")
+  n=$(printf '%s' "$raw" | grep -c .)
+  [ "$n" -eq 1 ] || { echo "found $n single-line banned='...' assignments in $f, expected exactly one: renamed, requoted, wrapped, or the gate now has more than one list, and summing them silently is how a partial list reads as the whole one" >&2; return 1; }
+  b=$(printf '%s' "$raw" | sed "s/banned='//; s/'$//")
   echo "alternates        $(printf '%s' "$b" | tr '|' '\n' | grep -c .)"
   echo "distinct products $(printf '%s' "$b" | tr '|' '\n' | tr -d ' -' | sort -u | grep -c .)"
 }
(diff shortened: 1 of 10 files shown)


mosamlife:
Closing unmerged. The owner has decided to remove `scripts/claude/` entirely rather than keep fixing it, and I agree with the reasoning.

`bash-guard.sh` decides what a command will write by parsing shell text. That is undecidable, not merely hard: `eval`, `bash -c`, a path built at run time, a variable holding a glob. Four rounds went by, each closed real bypasses that had been proved destructive by execution, and each following round found more. This PR is round four. It is better than round three and there is no reason to believe it is the last one.

The sharper argument against it is what the guards actually caught. Over a full day of work the real findings were a privilege escalation, a check-then-act race that could leave an organisation with no owner, an unclamped API key role, an inverted claim in a PR body, and several stale hard-coded counts. Every one came from an agent or a review bot attacking the work. The shell guards caught none of them.

So the review process is what is working, and it is being kept: `CLAUDE.md`, `review.md`, the specialist agent definitions, and adversarial verification of anything risky.

`.githooks/pre-push` is being kept as well, and it is worth saying why it is different in kind. Git hands it already-resolved refs, so `eval`, a `cd`, quoting and exotic refspecs are gone before it runs. It is not guessing. Three attack rounds threw 13 shell tricks and 11 push forms at it without a bypass, which the text matcher never managed.

The work in this PR is not lost — the branch stays, and the reproductions and mutation tables in the commit messages are the record of what these guards did and did not do.

PR #423, opened 2026-08-14
build: move the Go toolchain to 1.26.6

mosamlife:
`main` is red with no code change. The same commit `500b445d` was green at 19:03 and failed at 04:08
the next morning, because `govulncheck` reads a live advisory database. `.claude/rules/ci-and-build-logic.md`
already names this: *"it can redden main with zero code change and a green PR can fail on merge. Fix
the dependency. Never tag a red commit."*

Seven vulnerabilities, all standard library, every one `Found in: …@go1.26.5` / `Fixed in: …@go1.26.6`.
Reachable in this codebase through `blobstore.Store.EnsureBucket` to `s3.Client.HeadBucket`
(`encoding/xml`), `pgxpool.Pool.Close` (`encoding/asn1`), and `http.Client.Do` (`golang.org/x/net/idna`).
Recursion-depth and input-validation class, not remote code execution.

## The second defect, which is why this is more than a version bump

CI pinned `go-version: "1.26.5"` exactly while `infra/Dockerfile.api` built `FROM golang:1.26`, a
floating tag. Those can differ in either direction and nothing noticed — CI could pass on a toolchain
the shipped image never used. That floating tag has since moved on its own: `docker run --rm golang:1.26
go version` now prints `go1.26.6`, so the image's compiler changed without anyone deciding it.

Every declaration is now an exact pin, and `scripts/check-go-toolchain.sh` (with a test file, 11 cases)
asserts they are all identical and none floats. Six declarations across `.github/` and `infra/`,
including `infra/docker-compose.dev.yml`, which was a fourth floating tag nobody had listed — a grep
for `1\.26\.[0-9]` structurally cannot see a floating tag, which is how it stayed hidden.

`apps/api/go.mod` deliberately stays at `go 1.26.3`. That is the language floor for anyone building
from source, not the compiler that ships; raising it per patch release pushes a floor onto consumers
for no security gain. The guard asserts floor ≤ toolchain rather than equality.

## Verified

```
$ GOTOOLCHAIN=go1.26.6 govulncheck ./cmd/wpmgr/...
No vulnerabilities found.
Your code is affected by 0 vulnerabilities.
EXIT=0                       (advisory IDs in output: 0)

$ bash scripts/check-go-toolchain.sh          EXIT=0   (6 declarations found)
$ <revert one Dockerfile to golang:1.26>      EXIT=1
  FAIL: infra/Dockerfile.api:15 uses base image golang:1.26 — a floating tag.
$ <restore>                                   EXIT=0
```

The guard was reverted against the real repository rather than a synthetic fixture, and an empty tree
exits 1 — finding zero declarations is a failure, not a pass. The guard also went red on its own first
run against a genuine bug in itself: it read a comment as a declaration, because the `#` test ran
against `grep -n` output that is prefixed `file:line:` and so never starts with `#`. Both scans had it;
both now have regression cases.

## Not done, stated plainly

**This does not fix production.** The deployed binary is still `go1.26.5` — read out of the published
image, not inferred:

```
$ docker cp <container>:/usr/local/bin/wpmgr ./wpmgr && go version -m ./wpmgr
./wpmgr: go1.26.5
```

Closing it needs a rebuild and redeploy of both layers, and the published GHCR images self-hosters pull
carry the same toolchain. That is a release, not a merge.

`go test ./...` did not complete: 69 packages `ok`, 0 `FAIL`, then it stalled on `apps/api/tests`
spinning testcontainers, and was stopped at a bounded four minutes. `ci.yml` excludes that package by
name, so it does not run there either. Reported incomplete rather than implying coverage.

Residual advisories: 6, in modules required but not called by this code. Unreachable, pre-existing, and
not addressed by this bump.

## Findings, reported not fixed

`infra/Dockerfile.web` and `infra/Dockerfile.marketing` still float on `node:22-alpine` and
`nginx:1.27-alpine`. `node:22` moves minor and patch silently, and the marketing image ships only via
cloudbuild, so `release.yml` never sees it. The new guard is Go-only by design and does not cover them.

## Summary by CodeRabbit

* **Chores**
  * Updated Go toolchains to version 1.26.6 across CI, development containers, and build images.
  * Added automated checks to detect missing, inconsistent, or floating Go toolchain versions.
* **Tests**
  * Added regression coverage for valid and invalid toolchain configurations, including version compatibility checks.
* **Bug Fixes**
  * Improved security checks by enforcing consistency between workflow and container Go versions.

### DIFF
--- a/.github/workflows/api-integration.yml
+++ b/.github/workflows/api-integration.yml
@@ -187,7 +187,7 @@ jobs:
 
       - uses: actions/setup-go@v5
         with:
-          go-version: "1.26.5"
+          go-version: "1.26.6"
           cache-dependency-path: apps/api/go.sum
 
       # The tests build their containers with testcontainers-go talking to the
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
 
       - uses: actions/setup-go@v5
         with:
-          go-version: "1.26.5"
+          go-version: "1.26.6"
           cache-dependency-path: apps/api/go.sum
 
       # Build the api module on its own (workspace disabled) so it does not
@@ -326,9 +326,22 @@ jobs:
       - name: Version surface guard (openapi, changelog page, badge, agent, install pins)
         run: scripts/check-version-surfaces.sh
 
+      # Which Go toolchain compiles the shipped binaries. This job pinned
+      # go-version 1.26.5 while infra/Dockerfile.api built FROM golang:1.26, a
+      # floating tag — so the govulncheck below cleared a toolchain that was not
+      # necessarily the one in the image. Nothing detected that; the only way to
+      # learn the deployed toolchain was `go version -m` on a binary pulled out
+      # of the published image. Self-test first, same as above: a regressed
+      # guard is the more useful failure to read.
+      - name: Go toolchain guard self-test
+        run: scripts/check-go-toolchain_test.sh
+
+      - name: Go toolchain guard (workflow pins vs container base images)
+        run: scripts/check-go-toolchain.sh
+
       - uses: actions/setup-go@v5
         with:
-          go-version: "1.26.5"
+          go-version: "1.26.6"
           cache-dependency-path: apps/api/go.sum
 
       # govulncheck does call-graph reachability analysis, so it only fails on
--- a/Makefile
+++ b/Makefile
@@ -123,6 +123,18 @@ check-versions: ## Check every version-naming surface (docs, marketing, agent)
 check-versions-test: ## Run the version surface guard's regression suite
 	scripts/check-version-surfaces_test.sh
 
+# Which Go toolchain compiles the shipped binaries. CI pinned 1.26.5 while
+# infra/Dockerfile.api built FROM golang:1.26 (floating), so the tested
+# toolchain and the shipped one could differ and nothing noticed. This is the
+# check that makes them unable to disagree.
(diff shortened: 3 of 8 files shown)


mosamlife:
Superseded. Reopening as a smaller change that keeps the version pins and drops the guard this PR also introduced.

The bump itself was correct and is being kept: `govulncheck` went from exit 3 with seven standard-library advisories to exit 0 with none, and the Dockerfiles move from a floating `golang:1.26` to an exact pin, which is what stops CI and the shipped image from silently disagreeing about the compiler.

What is being dropped is `scripts/check-go-toolchain.sh`. It could be confidently wrong in two directions, both reproduced:

Replacing the Go base image with `alpine` in all three container declarations left it reporting `check-go-toolchain: OK`, exit 0 — the emptiness check counted declarations globally, so three surviving workflow pins masked the complete absence of the container half.

And with a version repeated inside a trailing comment, the greedy anchor read the last occurrence, so it reported agreement on `1.26.5` — a version present in no declaration in that tree. It invented a consensus rather than finding one.

Both were fixed on a follow-up branch. Neither is the reason for closing. The reason is that the guard is a third layer over a problem the first two already solve: the exact pins make the toolchain a declared fact, and `govulncheck` reads a live advisory feed and is what caught this one, unprompted, with no shell written. Adding 180 lines of shell to check a version string is the same instinct that produced the harness in #421, on the same day it was removed for having exactly this shape of defect.

The work is not lost; the branch stays. The replacement PR carries the pins alone.

Issue #414, opened 2026-08-12
Pause monitoring


iacopoincerpi:
**Describe the bug**: pause monitor is programmed to do what? Nothing happens.

**To reproduce**: pause monitoring on one site, nothing happens.

**Expected behavior**: pause performance check, uptime check, updates check...

**Environment**: 061.133, self host

<img width="636" height="204" alt="Image" src="https://github.com/user-attachments/assets/aeef5b7f-4630-4f95-a4fc-b12057785bed" />

infoproxa-oss:
That's never been implemented :)
There's a small pop-up that lets you know.

<img width="618" height="102" alt="Image" src="https://github.com/user-attachments/assets/0a637067-77aa-4756-8075-5a9de3010b2b" />

mosamlife:
Fixed in **0.61.137**, now deployed and published.

You were right that it did nothing. Pause wrote a flag and no scheduler read it, so it was a button with nothing behind it.

**What pausing now stops:** uptime checks, uptime alerts, the weekly screenshot fanout, and scheduled vulnerability rescans.

**What it deliberately does not stop:** backups, the WP-Cron kick that WordPress needs for its own scheduled work, connection checks, RUM, and retention.

**One thing you asked for that is not there, and I want to be straight about it rather than let you find out later.** You listed update checks. Those are not paused, because there is no scheduled update check to pause — inventory refresh runs only when someone clicks Refresh, or right after an update completes. The same is true of the Security scan button. So there was nothing running in the background to stop, and gating them on pause would have broken the one thing that should always work: anything you click yourself. If you are seeing update checks fire on a schedule in your install, that is a separate bug and worth its own issue — I would want to see it.

Pause is reachable from a site's row menu, from the site's own page, or in bulk from the sites list. It takes a reason and an optional resume time; resuming happens by hand or automatically at that time, exactly once, and both are recorded in the audit log.

**One behaviour change worth knowing about, since you self-host and read the reports.** The monthly report's uptime section now goes by whether a pause actually overlapped the period being reported on, rather than by whether the site happens to be paused when the report runs. Previously, pausing a site today could blank out last month's fully-measured uptime. Now a period a pause never touched is shown in full, and a period a pause partly covered says how many hours went unmeasured instead of looking like a complete month.

**To get it**, pull `v0.61.137` — the api, web and media-encoder images are published and I have confirmed all three are pullable at that tag.

Leaving this open until you have had a chance to confirm it behaves as you expected on your install.

[...]

PR #429, opened 2026-08-15
feat(db): add the monitoring-pause columns to sites (GH #414, phase 1 of 5)

mosamlife:
Refs GH #414. Phase 1 of 5, and it deliberately changes nothing.

## Why this exists

The bulk menu offers **Pause monitoring on N sites**. Selecting it runs:

```js
// apps/web/src/routes/_authed/sites/index.tsx:789
toast.info(`Pausing monitoring on ${selection.count} sites lands in Sprint 4`);
```

A maintainer confirmed on the issue: *"That's never been implemented :)"*. There is no pause concept
anywhere in the product. This adds the place to record one.

## What this PR is

Four columns on `sites`, a CHECK, one index, and the regenerated sqlc output. **No Go reads them.**
Nothing pauses, nothing is scheduled differently, no endpoint exposes them. The routes are a separate
PR, and the scheduler predicates come later still.

That order is the point: the risky part of this feature is teaching a dozen workers a new predicate,
and it should land on a foundation that has already been reviewed rather than alongside one.

```
monitoring_paused_at     timestamptz            -- NULL means active; the flag and the since-when
monitoring_paused_by     uuid                   -- attribution, ON DELETE SET NULL
monitoring_paused_reason text NOT NULL DEFAULT ''
monitoring_resume_at     timestamptz            -- optional auto-resume, NULL means until resumed
```

## Why a column and not a connection state

`sites.connection_state` is a state machine (ADR-041, `internal/site/connection.go:20-36`):
`pending_enrollment → connected → degraded → disconnected → revoked → archived`. Every one of those
describes **whether the agent is reachable**. Pause describes **whether we choose to act**. They are
orthogonal: a connected site can be paused, and a paused site can lose its agent. Folding pause into
that enum makes both facts unrepresentable at once.

## The three decisions, each defended rather than defaulted

**`ON DELETE SET NULL` on `monitoring_paused_by`.** Six `REFERENCES users` exist in the migrations;
the four attribution columns are unanimously `SET NULL` and the only `CASCADE`s are ownership rows.
`CASCADE` here would delete the **site**. `RESTRICT` would let an old pause block account deletion.
Note the database cannot enforce that the pauser belongs to the site's tenant — there is no
`users (id, tenant_id)` key — so the service must set it from the authenticated actor, never from
request input.

**No index on `monitoring_paused_at IS NULL`.** That predicate matches nearly every row, so it is a
full-size index wearing a `WHERE`, paid on every write to the most-written table here. And nothing
could use it: the fleet enumerations in `db/query/sites.sql` are uncapped sequential scans with no
index on `enrolled_at` at all. This is inferred from query shape, not measured against production
data; phase 2 should re-`EXPLAIN` if the fleet grows an order of magnitude. What did ship is
`sites_monitoring_resume_due_idx`, near-empty by construction, on the only genuinely new scheduled
query.

**The CHECK lives in the schema.** `monitoring_resume_at IS NULL OR monitoring_paused_at IS NOT NULL`.
A dangling resume instant is not cosmetic: it is a row the phase-2 sweep would pick up and act on. It
deliberately does not tie `monitoring_paused_by` to the pause, because the FK's own `SET NULL` would
violate that on every user deletion — those two decisions are coupled and only this combination is
consistent.

## RLS

No new policy, and none needed: RLS is row-level, so new columns on an admitted row inherit it.
Asserted live rather than assumed — `sites: ENABLE=true FORCE=true sites_site_scope=RESTRICTIVE`.

## Proof

Five proofs run through `site.Repo` and `InTenantTx` as the non-superuser `wpmgr_app` role, each on a
fresh container running the same migrate path `main()` runs at boot:

```
monitoring_paused_at=<nil> monitoring_paused_by=<nil> monitoring_paused_reason="" monitoring_resume_at=<nil>
CHECK fired as intended: 23514 sites_monitoring_resume_requires_pause_check
after user delete: paused_at=<set> paused_by=<nil> reason="noisy staging box"
indexdef: CREATE INDEX sites_monitoring_resume_due_idx ON public.sites USING btree (monitoring_resume_at)
          WHERE ((monitoring_resume_at IS NOT NULL) AND (monitoring_paused_at IS NOT NULL))
```

`go test -timeout 45m ./tests/...` → `ok … 587.776s`. sqlc regenerated with v1.31.1, matching the tree
stamp; `git diff --quiet -- internal/db/sqlc/` exits 0 after a re-run, so nothing was hand-edited.

## Reviewed

A security review and an independent adversarial review both ran against this work. Their findings
were against the **routes**, which are held back in a separate branch — cross-tenant reach was proved
closed by neutralising the explicit `tenant_id` predicate and confirming the boundary still held on
RLS alone.

One pre-existing finding worth recording, not fixed here: `sites_site_scope` appears in
`apps/api/db/schema.sql` only in comments, at lines 348 and 365. The policy itself is in the
migrations. Anyone grepping `schema.sql` to decide whether `sites` is site-scoped would conclude it is
not.

## Not done

Nothing is deployed. The routes, the scheduler predicates, the auto-resume sweeper and the interface
are phases 2 to 5.

## Summary by CodeRabbit

* **New Features**
  * Added support for pausing site monitoring with a recorded timestamp, reason, pausing user, and optional scheduled resume time.
  * Preserved pause details when viewing, listing, updating, enrolling, restoring, or checking site connections.
  * Added safeguards to ensure scheduled resumes apply only to paused sites.

* **Bug Fixes**
  * Improved data integrity and attribution handling for monitoring pause records.


mosamlife:
Superseded by #432, which carries the whole of GH #414 end to end.

This was phase 1 of five. The later phases stacked onto this branch rather than sitting beside it, so every commit here is already contained in #432 — verified with `git merge-base --is-ancestor`. Merging this separately would land the same commits twice.

Closing rather than merging. The branch is left in place, so the phase-1 review history stays reachable.

PR #430, opened 2026-08-15
feat(api): pause and resume monitoring on sites (GH #414, phase 1 of 5)

mosamlife:
Refs GH #414. Contains #429 (the migration) and adds the two routes. **Still changes no scheduling
behaviour** — nothing reads the flag yet, so nothing pauses.

Both a security review and an independent adversarial review ran against the first draft of this and
returned fix-first. Everything below is either what they confirmed safe or what they made me change.

## The routes

```
POST /api/v1/sites/monitoring/pause    site_ids []uuid, reason string?, resume_at timestamp?
POST /api/v1/sites/monitoring/resume   site_ids []uuid
```

Bulk-shaped from the start, because the menu is a bulk action and a single-site route would have to be
written twice. Idempotent: re-pausing does **not** overwrite `paused_at`, `paused_by` or the reason, so
a client that retries on timeout cannot erase what someone typed. One audit event per site, not per
request, so auditing a single site finds it.

## What the reviews found, and what changed

**The site-scope policy was inert on both write paths.** The repo hard-coded `InTenantTxAsUser`, so
`app.site_scope` was never set and the RESTRICTIVE `sites_site_scope` policy never engaged. A
site-scoped collaborator was stopped by exactly one Go line. Both paths now acquire their transaction
through `pool.RunTenantTx`, which dispatches to `InScopedTenantTx` — the same pattern
`internal/site/repo.go:List` already uses. Two independent layers, not one `if`.

Proved by execution rather than reading: a test calls the repo **directly** with a site-scoped
principal, no handler and therefore no `CanAccessSite` gate anywhere in the path. The out-of-scope site
returns 0 rows and stays unpaused; the control in the same subtest confirms that principal still pauses
its own site.

**The "no observable change" claim leaked, and I had it backwards.** A pause wrote `updated_at`, which
is served on `GET /sites/{id}`. My instinct was to accept it and correct the claim. The better answer
was to remove the write, because `sites.updated_at` is also served as `as_of` on
`GET /sites/{id}/updates` — the freshness stamp for the plugin and theme inventory. Bumping it on a
pause claims an inventory refresh that never happened, which is precisely the kind of lie this feature
must not tell. Dropped from pause and resume both.

**An 888 KB junk request returned a 7.5 MB body** with 100,001 result entries, against a documented
`maxItems: 200`, because the cap was checked after partitioning. A `MaxBytesReader` now refuses an
oversized body before it is parsed and the cap is enforced on the raw array. Measured: a 3,500,014-byte
request now returns **422 in 66 bytes**.

**The same mistake produced two different errors.** A bad `resume_at` gave `resume_at_in_past` to one
caller and `site_ids_required` to another, and an all-rejected list returned a 422 instead of the
per-site report the route promises. Validation now runs before the empty check, and the comment that
claimed to prevent exactly this divergence is replaced with what the code actually does.

**Archived and revoked sites are refused**, reported as `site_archived` / `site_revoked`. All three
lifecycle states paused happily before, and `archived` is the soft-delete state hidden from the default
sites list, so such a site could never have been resumed from the interface. Resume is deliberately not
gated, or archiving a paused site would strand it.

## Ten guards that could be deleted with every test still green

The adversary removed the `CanAccessSite` gate, the dedup, both size caps, `tenant_required`,
`invalid_resume_at`, the API-key `nullableUUID` path, both `FOR UPDATE` clauses and the audit metadata
— all at once — and all nine subtests plus the contract package stayed green.

Twenty guards now have a test each, every one shown red then green:

```
 1 CanAccessSite gate          11 Principal threading
 2 dedup                       12 validation-before-empty ordering
 3 200-cap (handler)           13 empty-list report
 4 200-cap (service)           14 RunTenantTx -> InTenantTxAsUser
 5 reason-length cap           15 FOR UPDATE (pause)
 6 tenant_required             16 FOR UPDATE (resume)
 7 invalid_resume_at           17 audit metadata
 8 nullableUUID                18 archived/revoked predicate
 9 MaxBytesReader              19 nullableUUID FK, end to end
10 principal_required          20 updated_at left alone
```

The mutation harness carries two controls on itself: editing a comment leaves the suite green, so it
does not over-fire; and a pattern matching nothing reports MUTATION-NOT-APPLIED rather than scoring a
pass. Both matter, because this repository has repeatedly shipped assertions that pass regardless of
the code — including inside the fixes for that problem.

## Confirmed safe, by mutation rather than by reading

Cross-tenant reach is closed at the layer below the application. The security reviewer **neutralised**
the explicit `tenant_id = $1` predicate in the pause SQL, replacing it with an always-true bound
predicate, and the boundary still held on RLS `sites_tenant_isolation` alone. Tenant A naming only
tenant B's site returns `{"ok":false,"detail":"site_not_found"}` and the victim row, read back in its
own scope, is untouched. No existence oracle: a foreign-but-real id and a never-existed uuid produce
byte-identical responses.

Concurrent pauses were run eight rounds with two operators and different reasons: exactly one request
reported changed, one audit row, and `monitoring_paused_by` always the winner.

## One thing broken and fixed in the same branch

Wrapping the JSON bind in a helper made `tests/contract/openapi_request_body_test.go` unable to find
it — that test locates a request body by finding the `ShouldBindJSON` call **by name** inside the
handler function, so the whole field-drift check went red. The `MaxBytesReader` stays; the bind is back
inline in each handler, and that trap is now written down beside it.

## Open

`make test-integration` reported one failure, `TestAutologinMintInjectsRealStoredDefault`, which passes
in isolation and touches no file in this diff. A full-suite run on `main` is in flight to establish
whether it is pre-existing; **this PR should not merge until that answers**, because "passes in
isolation" is not evidence.

For `apps/web`: two new `detail` values, `site_archived` and `site_revoked`, plus two new error codes
`request_too_large` and `principal_required`. The bulk-action UI should name them rather than treating
an unknown detail as a generic failure.

Phases 2 to 5 — the scheduler predicates, the interface, exclusions and the auto-resume worker — are
separate.

## Summary by CodeRabbit

* **New Features**
  * Added bulk site-monitoring pause and resume actions.
  * Pause requests can include a reason and optional scheduled resume time.
  * Responses show per-site outcomes, validation or access issues, current pause details, and total changes.
  * Repeating pause or resume requests is safely handled without unintended duplicate changes.
  * Added audit records for monitoring pauses and resumes.
* **Bug Fixes**
  * Improved validation for unauthorized, archived, revoked, duplicate, oversized, and invalid requests.
  * Added safeguards to prevent inconsistent pause and resume states.


mosamlife:
Superseded by #432, which carries the whole of GH #414 end to end.

This was phase 1 of five. The later phases stacked onto this branch rather than sitting beside it, so every commit here is already contained in #432 — verified with `git merge-base --is-ancestor`. Merging this separately would land the same commits twice.

Closing rather than merging. The branch is left in place, so the phase-1 review history stays reachable.

PR #431, opened 2026-08-15
test(api): make container setup failures say they are setup failures

mosamlife:
`TestAutologinMintInjectsRealStoredDefault` failed once in a full-package run and has passed every run
since, including a full run of the identical code. **This PR does not fix that**, because it was never
reproduced. It makes the next occurrence diagnosable in one glance instead of starting the same
investigation again.

## What was ruled out, by measurement

An investigation went after the obvious causes and disproved all of them:

- **Not shared database state.** Every test starts its own Postgres and Redis. `grep -c "startPostgres(t)"`
  returns 12 in `autologin_integration_test.go` and 3 in `autologin_policy_integration_test.go` — 15
  calls for 15 test functions, no `sync.Once`, no cached package pool.
- **Not shared limiter state.** `autologin.NewMemoryLimiter()` is constructed inline per test and is a
  plain in-process map, not Redis-backed.
- **Not parallelism.** `grep -rn "t.Parallel()" apps/api/tests/` returns two hits, both in
  `scan_plugin_checksums_regression_test.go`.
- **Not clock skew, today.** `docker run --rm alpine date +%s` and host `date +%s` both returned
  `1786792156`. Zero.

## What is actually wrong

`startPostgres` turned a container **start** failure into `t.Skipf`, but turned connection-string,
migrate, connect and role-provisioning failures into `t.Fatalf` with a bare `%v`. `startRedis` and
`connectAdmin` had the same shape.

So during a full run — hundreds of containers over about ten minutes — an infrastructure failure
surfaced as *this test asserting wrongly*. That fits the evidence exactly: fails in a full run, never
alone.

Every setup stage now fails with a message that cannot be mistaken for an assertion:

```
SETUP FAILURE (infrastructure, not the test's own assertion) at stage="postgres: run embedded migrations": ensure schema_migrations: closed pool
```

Fail rather than skip, applied consistently to every stage: skipping would hide a real breakage across
the whole suite, and the point here is classification, not silence. Container-start keeps its existing
skip. `connectAdmin`'s "no DSN recorded" is caller misuse rather than infrastructure and says so
separately.

## Proved, by forcing each stage

```
unreachable DSN     → stage="postgres: connect as bootstrap superuser": ping: ... connection refused
pre-closed pool     → stage="postgres: run embedded migrations": ensure schema_migrations: closed pool
bad redis port      → stage="redis: resolve mapped port": port "9999/tcp" not found
```

Each was forced, captured, and reverted; `git diff` was clean before committing. Does not over-fire:
five representative tests still pass, `grep -c '^--- PASS'` → 5.

## One latent fuse, documented not fixed

Autologin writes `expires_at` from the **host** clock (`s.clock.Now().UTC().Add(JWTTTL)`, 60 seconds)
while `db/query/autologin.sql:33` gates on `expires_at > now()`, which is the **container's** clock.
Skew is zero today, so this is not a live bug and no production code or query was changed. A comment
now names the cross-clock comparison and the 60-second budget, so whoever meets an intermittent
`Consume` failure finds the explanation rather than rediscovering it. Every autologin consume test sits
on that same fuse.

Also left alone deliberately: `internal/autologin/service.go:188` swallows the audit-recorder error, so
a failed audit write surfaces as `no rows in result set` with no cause. That is a real diagnosability
gap, and it is also intentional best-effort behaviour in production. Changing production behaviour to
improve a test message is the wrong trade; recorded here instead.

## Scope

Two test files, 48 insertions, 11 deletions. No production code. Independent of GH #414.

## Summary by CodeRabbit

* **Tests**
  * Improved integration-test setup error handling to clearly distinguish infrastructure failures from test assertion failures.
  * Added clearer diagnostics for container startup, database connections, migrations, role provisioning, and reconnection issues.
  * Tests now skip cleanly when Docker is unavailable.
  * Documented potential clock-skew effects on autologin token expiration tests.

### DIFF
--- a/apps/api/tests/autologin_integration_test.go
+++ b/apps/api/tests/autologin_integration_test.go
@@ -40,6 +40,9 @@ import (
 func startRedis(t *testing.T) (*redis.Pool, string) {
 	t.Helper()
 	ctx := context.Background()
+
+	skipIfDockerUnavailable(t, ctx, "redis")
+
 	req := testcontainers.ContainerRequest{
 		Image:        "redis:7-alpine",
 		ExposedPorts: []string{"6379/tcp"},
@@ -50,7 +53,7 @@ func startRedis(t *testing.T) (*redis.Pool, string) {
 		Started:          true,
 	})
 	if err != nil {
-		setupSkipf(t, err, "redis: container start (docker unavailable?)")
+		setupFatalf(t, err, "redis: container start")
 	}
 	t.Cleanup(func() { _ = c.Terminate(ctx) })
 	host, err := c.Host(ctx)
@@ -102,16 +105,17 @@ func newAutologinSigner(t *testing.T) *agentcmd.Signer {
 // Recorder.
 //
 // Cross-clock note: domain.SystemClock{} below is the HOST's clock — Mint
-// computes expires_at as s.clock.Now().UTC().Add(autologin.JWTTTL) (60s,
-// internal/autologin/model.go) on the host, but Consume's gate is
-// `expires_at > now()` evaluated by the testcontainer's OWN Postgres clock
-// (db/query/autologin.sql, ConsumeAutologinToken). A Consume call in these
-// tests is therefore only guaranteed to pass if host and container clocks
-// agree within the 60s budget minus whatever wall time elapsed between Mint
-// and Consume. Measured skew here is currently zero, so this is not a live
-// bug — but an intermittent Consume failure ("nonce_unavailable") under load
-// is where a real skew would show up first, and this comment is the
-// explanation to find instead of re-deriving it.
+// computes expires_at as s.clock.Now().UTC().Add(autologin.JWTTTL)
+// (internal/autologin/model.go is the source of truth for the TTL value) on
+// the host, but Consume's gate is `expires_at > now()` evaluated by the
+// testcontainer's OWN Postgres clock (db/query/autologin.sql,
+// ConsumeAutologinToken). A Consume call in these tests is therefore only
+// guaranteed to pass if host and container clocks agree within the JWTTTL
+// budget minus whatever wall time elapsed between Mint and Consume. Measured
+// skew here is currently zero, so this is not a live bug — but an
+// intermittent Consume failure ("nonce_unavailable") under load is where a
+// real skew would show up first, and this comment is the explanation to find
+// instead of re-deriving it.
 func buildAutologinService(t *testing.T, pool *db.Pool, redisPool *redis.Pool, siteSvc *site.Service, cfg autologin.Config) (*autologin.Service, *audit.Recorder) {
 	t.Helper()
 	rec := audit.NewRecorder(pool, domain.SystemClock{})
--- a/apps/api/tests/autologin_integration_test.go
+++ b/apps/api/tests/autologin_integration_test.go
@@ -52,10 +52,19 @@ func startRedis(t *testing.T) (*redis.Pool, string) {
 		ContainerRequest: req,
 		Started:          true,
 	})
-	if err != nil {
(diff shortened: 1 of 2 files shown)


mosamlife:
Picking this up because today it would have paid for itself twice — and the open Major
finding is correct, so it should not merge as-is.

## The Major finding is real

`rls_integration_test.go:69` routes a container **start** failure to `setupSkipf`;
every later stage (`:75` connection string, `:81` superuser connect, `:84` migrations,
`:100` role provisioning, `:108` connect as `wpmgr_app`) correctly routes to
`setupFatalf`.

A skipped test prints `--- SKIP` and the package still prints `ok`. And this suite is
**not run by CI** — `.github/workflows/ci.yml:53` and `:95` both `grep -v` the
`apps/api/tests` package, which is why `make test-integration` exists as a manual
pre-merge step. So the only environment that runs it is one where a skip reads as a
pass.

That is this project's signature defect sitting inside the PR written to eliminate it:
Docker hiccups, every test skips, `make test-integration` says `ok`, and the merge
proceeds on a suite that asserted nothing.

The distinction worth drawing is between **Docker genuinely absent** — a contributor
without it, where skipping is the right UX — and **Docker present but the container
failed to start**, which is a real failure and must be red. Those are different
conditions and only the first deserves a skip. Failing closed by default, with the skip
as the explicit opt-in, is the safer direction given nothing downstream re-checks.

## The other finding is straightforwardly right

`autologin_integration_test.go:104-111` hard-codes `60s` for a TTL whose source of truth
is `autologin.JWTTTL`. Deriving it from the constant costs nothing, and this repo has a
standing rule against hard-coded derived quantities precisely because they drift
silently.

## Why this matters beyond its own diff

Twice today an investigation lost time to exactly the ambiguity this PR targets. A
reviewer chasing five simultaneous test failures could not attribute them because a
`grep '^--- FAIL'` filter dropped the assertion text beneath each failure line, leaving
a container error and an assertion error indistinguishable. I then reached for test
*duration* as a proxy for "died in setup" — 1.3-1.5s against a 4.5s norm — and that
inference was wrong: a later run showed passing tests at 1.57s, because per-test
duration tracks container warmth, not outcome.

Neither of those detours happens if setup failures announce themselves. That is the
value here, and it is worth the two fixes above rather than merging around them.

Branch is `BEHIND`; it needs updating against `main` as well.

mosamlife:
Resolved the outdated `60s` thread. Verified rather than dismissed:

```
git show origin/test/setup-failures-are-legible:apps/api/tests/autologin_integration_test.go | grep -c "60s"     -> 0
git show origin/test/setup-failures-are-legible:apps/api/tests/autologin_integration_test.go | grep -c "JWTTTL"  -> 2
```

The comment now names `autologin.JWTTTL` and points at `internal/autologin/model.go` as the
source of truth instead of stating a number it cannot guarantee.

Worth recording what that comment describes, because it stopped being hypothetical today:
it warns that `domain.SystemClock{}` is the HOST's clock while `Consume`'s gate is
`expires_at > now()` on the **testcontainer's** Postgres clock, and predicts the symptom as
an intermittent failure under load.

That is the same two-clock defect that produced a real failure in the GH #463 dispatch tests
during a full-suite run — the Docker VM clock lags the host across a host sleep until it
resyncs, so a row seeded one minute in the host's past sat in the database's future. Fixed
there in #498 by anchoring every seed to `SELECT now()` through the same pool. The comment's
"measured skew here is currently zero, so this is not a live bug" is true only between sleeps.

Not changing it here — this PR is about setup-failure legibility, and the autologin path
mints and consumes within one test rather than seeding a timestamp for a later scan, so the
exposure is smaller. Flagging it so the connection is on the record.

PR #432, opened 2026-08-16
feat: pause and resume monitoring on sites (GH #414)

### LINKED ISSUE #414: Pause monitoring

mosamlife:
Closes #414.

An operator can pause monitoring on a site — from its row menu, from its own
page, or in bulk from the sites list. The paused state is visible in all three
places, with the reason and the scheduled resume time.

## What stops

Uptime probes, uptime alerts, the weekly screenshot fanout, and scheduled
vulnerability rescans.

## What does not stop

Backups, the WP-Cron kick (which is what boots WP-Cron, which runs the
backups), the connection sweep, RUM, retention — and anything a person clicks.
Pausing stops the schedule, never the operator. Each of these is pinned by a
test that fails if a future change makes it respect pause.

## Resuming

By hand, or automatically at a stored time, exactly once. Both are audited,
including the automatic path.

## Reporting

The monthly report's uptime section is keyed on whether a pause overlapped the
reporting window, not on whether the site happens to be paused when the report
runs. A window it measured is never discarded; a window with a gap is labelled
rather than presented as measurement; a failed history read says coverage is
unconfirmed instead of claiming full coverage.

## Review

Built in five phases. Each phase took a security-reviewer pass and an
independent adversarial pass; the whole range then took both again, plus three
further adversarial passes covering the areas earlier ones had not reached.
Findings from those passes — a badge that greyed out a live outage, a report
that discarded a measured month, assertions that could not fail, and four
unguarded UI paths — are fixed in this branch, each with its regression test
shown failing before passing.

The full integration suite passes locally; `ci.yml` does not run that package.

## Summary by CodeRabbit

* **New Features**
  * Added controls to pause and resume monitoring for individual or multiple sites, with reasons and optional scheduled resumption.
  * Added per-site operation results, validation feedback, and partial-success reporting.
  * Site views now show pause status, pause details, health-check timestamps, and updated uptime indicators.
  * Reports identify paused, partially covered, or unconfirmed monitoring periods.
  * Scheduled monitoring, alerts, screenshots, and vulnerability scans now respect paused sites.
  * Monitoring automatically resumes when scheduled pause times expire.

* **Bug Fixes**
  * Prevented paused sites from affecting alert counts, outage notifications, and uptime averages.


mosamlife:
## Bot review triage — all 13 findings settled

**Fixed (9).** Each carries a regression test shown failing before it passes.

| Finding | Verdict |
|---|---|
| Paused notice denies uptime the same document prints | Real. The notice checked current pause state, the section checked window overlap. Both renderers fixed to consult the same fact. |
| `TestPDFFullCoverageUnchanged` negative assertion vacuous | Real. It searched zlib-compressed output, so it could never match. Second unfirable assertion found in that file. |
| `t.Fatalf` from a non-test goroutine | Real, though one level below the flagged line — via a decode helper. Proved with a standalone repro that the test does not abort. |
| Count read but never asserted | Real. Fourth unfired assertion inside this feature's own tests. |
| `fireAppAggregate` doc/code mismatch | Real. Code moved to `affected != nil` to match the documented contract. |
| Pause reason bounded in bytes, not characters | Real. The message already said "500 characters" while enforcing 500 bytes — a third of the stated room in CJK. |
| `detail` enum missing `site_archived`/`site_revoked` | Real. Now pinned by a test that parses the handler's AST, so a future code added without a spec update fails. |
| `principal_required` asserts an unknowable cause | Real, and worse than reported: it fires on any bare 403, while this codebase returns 401 for expired sessions. |
| Refusal copy ungrammatical | Real — produced "the site is no longer exists" for 2 of 9 codes. Rewritten with a graceful fallback for unseen codes. |

**Not confirmed (1).** `uptime_pct` null before `toFixed`: traced end to end — the handler only sets the field when the pointer is non-nil, the generated type is `OptFloat64` rather than `OptNilFloat64`, and its encoder emits the key or omits it, with no path that writes `null`. The wire is number-or-absent. No change made.

**Previously refuted, restated with evidence (3).**

- *Generated files edited in place* — refuted by regeneration, not argument. All three generators re-run; `git status` clean on `internal/api/gen/`, `internal/db/sqlc/` and `packages/openapi-client/src/generated/`. Committed output is byte-identical to fresh generator output.
- *Code breaks against pre-migration schema* — `Migrate()` runs synchronously in `main.go` before `srv.Run`, so a process never serves traffic until it has applied every migration in its own binary. The reverse case (old replicas, new schema) is what m117 is written to survive: nullable or defaulted columns, idempotent, no reader or writer in phase 1.
- *Migration should use `NOT VALID` / `CONCURRENTLY`* (on #429) — refuted with measurements. The migration's first statement already takes `ACCESS EXCLUSIVE` and holds it to commit, so `NOT VALID` releases nothing; `CONCURRENTLY` is illegal inside the transaction the runner uses, which is the same evidence the finding cited. At 200k sites: constraint validation 22.6 ms, index build 14.8 ms, all 128 migrations 1.33 s.

One half of that last finding was right and is kept: the index predicate is logically redundant but load-bearing for the planner. `EXPLAIN` shows 13.8 ms on the index versus 395 ms seq-scanning 199,700 rows without it. Documented at both the query and the index, each pointing at the other.

Issue #463, opened 2026-08-19
Deferred dispatch for scheduled update runs


mosamlife:
## Summary

The update wizard accepts a "Schedule (optional)" time and the API stores it in
`update_runs.scheduled_at`, but nothing defers the work: tasks are enqueued
immediately on create, so a run scheduled for 02:00 executes as soon as it is
submitted. **This is a live broken promise in the UI, not only a roadmap gap** —
an operator picks a time, and the fleet updates now.

This is narrower than it appears. The control plane already has a working, tested
scheduler used by backups, client reports and email digests: periodic sweeper,
per-entity due time, advisory-lock single-flight, `FOR UPDATE SKIP LOCKED`
claiming, timezone-aware next-occurrence with DST coverage, per-site jitter.
**Nothing needs to be invented.** Update runs need to opt into deferral, and the
surrounding guards need to be taught that "waiting for a start time" is not "stuck".

## Current behaviour

- `apps/api/db/schema.sql:929-930` — the column exists and is documented.
- `apps/api/internal/update/handler.go:77-80`, `repo.go:150-159` — accepted and persisted.
- `apps/api/internal/update/service.go:200-207` — every task is enqueued immediately.
  `grep -n "scheduled_at\|ScheduledAt" apps/api/internal/update/service.go` returns
  nothing, against a control of 16 functions in that file.
- `apps/web/src/features/updates/update-wizard.tsx:441-453` — the UI offers the control.
- `README.md:393` records this accurately.

## Why the obvious fix is unsafe on its own

Deferring only the queue insert, leaving task rows `pending`, breaks two existing guards:

1. **The stale-task reaper** sweeps `update_tasks WHERE status IN ('pending','running')
   AND updated_at < now() - 45m` (`apps/api/db/query/updates.sql:298-302`, threshold
   `apps/api/internal/update/worker.go:1085`) every 10 minutes with `RunOnStart: true`
   (`apps/api/cmd/wpmgr/main.go:3395-3399`). Any run scheduled more than 45 minutes out
   would be marked **failed** before it fires, with the misleading reason "stale: task
   exceeded max runtime".
2. **The cross-run dedup index** `update_tasks_inflight_target_idx` covers the same
   status set (`apps/api/db/schema.sql:996-998`). A task left `pending` overnight holds
   the slot for that (site, target), so **scheduling tomorrow's update would block
   updating that plugin today** with `409 targets_in_flight`.

## Proposed approach

[...]

PR #490, opened 2026-08-19
fix(db): m119 — declare 'halted' and 'cancelled' in the update status contracts

### LINKED ISSUE #482: update_runs.status: 'halted' is written by the code but declared in no migration or contract comment
## What is wrong

`update_runs.status` carries a value the database contract does not declare.

`RunHalted = "halted"` (`apps/api/internal/update/model.go:31`) is written by the agent
self-update wave machine and has been for a long time — it predates GH #463 entirely.
But:

```
$ grep -ln "halted" apps/api/migrations/*.sql
(no output)
```

Empty, against a control that matches — `grep -lc "completed" apps/api/migrations/*.sql`
returns real files. So `halted` appears in **no migration at all**, and it is absent
from the run-status block in `apps/api/db/schema.sql`.

That matters because **m118's `COMMENT ON COLUMN` declares itself the contract** for
this column, and there is **no CHECK constraint** — the comment is the only thing
saying which values are legal. A value the code writes routinely is missing from it.

## Why it surfaced now

GH #463's cancel path writes `halted` deliberately: a cancelled run has no `cancelled`
run status to go to, because minting one would create a value `gen.UpdateRunStatus`
cannot render. So `halted` is now written by two different mechanisms — an agent wave
halt, and an operator cancelling a scheduled run — and the contract declares neither.

Anyone reading that comment to learn the legal values gets an incomplete list, and the
incompleteness is invisible: nothing rejects the write.

## Fix

A new migration ordinal amending the `COMMENT ON COLUMN` for `update_runs.status` to
include `halted`, with its meaning. m118 is applied and immutable, so this cannot be an
edit to it.

While there, check the same column comment against `model.go`'s full set — m118 added
`scheduled`, `dispatching` and `expired`, and this issue proves at least one value was
already missing before that, so the list is worth reconciling once rather than
patching per-discovery.

**Consider whether a CHECK constraint is warranted.** The recurring argument in this
codebase is that the comment is the contract because no constraint exists — which is
true and is exactly why the comment drifted unnoticed. A constraint would make the
next omission fail loudly at write time instead of silently. That is a real decision
with a migration cost and a lock consideration on a live table, so it belongs to
whoever picks this up, not to this issue.

## Done when

- The column comment lists every value `model.go` can write, verified by comparing the
  two rather than by reading either alone. Paste the comparison.
- If a CHECK constraint is added, it is proven to reject an unlisted value and to
  accept every listed one — a constraint that rejects nothing is the same defect one
  level up.

mosamlife:
Fixes #482.

## The gap

`RunHalted = "halted"` (`apps/api/internal/update/model.go:31`) has been written by the agent self-update wave machine since long before #463, and appeared in **no migration at all**:

```
$ grep -ln "halted" apps/api/migrations/*.sql      # before this PR
(no output)

$ grep -lc "completed" apps/api/migrations/*.sql | head -3
apps/api/migrations/20260531030000_m17_backup_schedule.sql
apps/api/migrations/20260531020000_m16_restore_runs.sql
apps/api/migrations/20260531110000_m23_media_optimizer.sql
```

The control matches real files, so the empty result is a real absence and not a broken pattern. Confirmed independently against the m118 commit itself:

```
$ git show ee2a1c6b:apps/api/migrations/20260820000000_m118_update_runs_agent_rls.sql | grep -c halted
0

$ grep -ln halted apps/api/migrations/*.sql       # after this PR
apps/api/migrations/20260821000000_m119_update_status_contract_reconcile.sql
```

Neither `update_runs.status` nor `update_tasks.status` has a CHECK constraint, so m118's `COMMENT ON COLUMN` is the **entire** contract — m118's DECISION 3 says so in as many words. A value missing from it is not a documentation gap; it is the absence of the only definition that exists.

#463 then gave `halted` a **second writer** — an operator cancelling a scheduled run — so it is now reached by two subsystems and was declared by neither:

| writer | mechanism |
|---|---|
| `update/agent_repo.go` `haltLocked()` | a wave gate refused to advance an agent self-update rollout |
| `update/cancel_repo.go` `CancelScheduledRun()` | an operator cancelled a scheduled run before it fired (#463) |

## New ordinal, not an edit

m118 has applied, and `migrate.go` sorts lexically and skips anything already in `schema_migrations`, so editing it would be a silent no-op that looked like a fix. m118's own closing note specified this case in advance: *"that is a NEW ordinal amending the comment - never an edit to this file."*

`20260821000000` is the next free ordinal across **every commit on every ref**, not merely `main`, because the ordinal is apply order and not commit order:

```
$ git log --all --name-only --format= -- apps/api/migrations/ \
    | grep -oE '2026[0-9]{10}_m[0-9]+' | sort -u | tail -4
20260817000000_m115
20260818000000_m116
20260819000000_m117
20260820000000_m118
```

## The reconciliation, in full

The whole list was compared rather than the one reported omission patched, because a contract short by one value for months is not evidence of one mistake — it is evidence that nothing compares the two sides. Both columns were short by exactly one:

**Against m118 (the applied contract) — the check fires:**

```
$ /tmp/reconcile-482.sh . apps/api/migrations/20260820000000_m118_update_runs_agent_rls.sql
update_runs.status   (code writes 7, contract declares 6)
  MISSING from the contract (code writes it, comment omits it): halted
update_tasks.status  (code writes 9, contract declares 8)
  MISSING from the contract (code writes it, comment omits it): cancelled

DRIFT DETECTED (see above).
exit=1
```

**Against m119 (this PR) — clean, in both directions:**

```
$ /tmp/reconcile-482.sh . apps/api/migrations/20260821000000_m119_update_status_contract_reconcile.sql
update_runs.status   (code writes 7, contract declares 7)
  OK - the two sets are equal in both directions.
update_tasks.status  (code writes 9, contract declares 9)
  OK - the two sets are equal in both directions.

RECONCILED: every value the code can write is declared, and nothing is
declared that the code cannot write.
exit=0
```

**And it refuses rather than passing when it matches nothing** — pointed at a migration with no such comment:

```
$ /tmp/reconcile-482.sh . apps/api/migrations/20260819000000_m117_site_monitoring_pause.sql
REFUSING update_runs.status : extracted code=7 declared=0 value(s).
  An empty side means the pattern or the path is wrong, not that the
  sets agree. Fix the extractor before concluding anything.
exit=2
```

The reverse direction matters as much as the forward one: a contract listing a value nothing writes sends the next reader looking for a state machine that does not exist.

## Idempotence, proved against a real Postgres

A scratch Postgres 16 was brought to m118 (the exact population needing converging), then m119 applied **twice**. `md5` of each column comment after each step:

| capture | `update_runs.status` | declares `halted` | `update_tasks.status` | declares `cancelled` |
|---|---|---|---|---|
| A — at m118 | `4e3bbb1b…` (1462 b) | **f** | `b6784001…` (1470 b) | **f** |
| B — after 1st m119 | `7601bfee…` (2859 b) | **t** | `3c633ed0…` (2487 b) | **t** |
| C — after 2nd m119 | `7601bfee…` (2859 b) | **t** | `3c633ed0…` (2487 b) | **t** |

B and C are byte-identical. Both applications exited 0.

`COMMENT ON COLUMN` **replaces** the comment wholesale rather than appending or merging, so it is idempotent by construction and needs no `DO`-block existence guard of the kind m118's policy and index needed — there is no "already exists" state to skip, only a value to set.

**Converge path: none required**, and that is not a formality. This corrects a comment, not data and not structure, and a comment is last-write-wins with no history. A database at m118 that now applies m119, and a fresh database that applies m118 then m119, reach an identical end state. No row, index, policy or constraint differs. Critically, there is no population carrying a wrong *value* to converge, because nothing ever wrote a value based on the incomplete list — the list was only ever read by humans.

## Scope

Comment-only. No table, policy, index, constraint or row is touched.

- `apps/api/db/schema.sql` is brought to the same end state as inline `--` comments, following this repo's convention: that file contains no `COMMENT ON` statements anywhere and m118 documented why introducing one would buy nothing.
- `atlas.sum` is untouched. It has been stale since m100, `migrate.go` does not verify it, and m118 did not touch it either.
- `sqlc generate` (v1.31.1, matching the tree stamp) produces **zero diff** — confirming this touches no column and no query. The generated tree was regenerated, not hand-edited.

`go build ./...`, `go vet ./...` and `go test ./internal/update/...` all pass.

## No CHECK constraint here — deliberately

See the review note below. It is the obvious response to a contract that drifted unnoticed for months, and it is deferred to its own decision, its own ordinal and its own review rather than smuggled in behind a comment change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented halted or expired update runs from being incorrectly changed to another status.
  * Preserved halted status when a run finishes, including after external shutdown conditions.
  * Avoided reporting or recording halt events when a halt request is refused.
  * Maintained normal completion behavior for active runs.
* **Documentation**
  * Clarified update run and task status lifecycles, terminal states, cancellation outcomes, and status transitions.
* **Tests**
  * Added coverage for halted-run persistence, refused halts, and standard completion flows.


mosamlife:
## Should `update_runs.status` / `update_tasks.status` get a CHECK constraint?

**Recommendation: no — and #482 is the argument against it, not for it.** Deferred to its own decision either way; nothing in this PR adds one.

### The case for is real

The comment is the contract *because* no constraint exists, and that is exactly why it drifted unnoticed for months. A constraint would move the failure from "silent, found by a human reading two files" to "loud, at write time".

### Why it would not have caught this, and would have made it worse

`halted` and `cancelled` are not typos. They are **correct, intended values written by correct code**. A CHECK added at m118 would have been enumerated from the same reading of `model.go` that produced the incomplete comment — the author who omitted `halted` from the comment omits it from the CHECK, because it is the same list typed twice in the same sitting.

The outcome then is not "the next omission fails loudly". It is:

> every wave-gate halt raises `23514` inside `haltLocked`, in the transaction that cancels pending tasks during an agent self-update rollout that has **already gone wrong**.

The kill switch breaks at the exact moment it is most needed, and the operator's cancel path (#463) breaks with it. Today an undeclared value costs a stale comment; with a CHECK built from that same incomplete list it costs the halt path. That is a strictly worse failure mode, and m118 predicted it in as many words — *"getting that list wrong turns a typo into a 23514 outage inside main()"*. #482 is the empirical demonstration that getting the list wrong is precisely what happens.

### What it would cost

- **Boot-time outage risk, fleet-wide.** `ADD CONSTRAINT … CHECK` takes ACCESS EXCLUSIVE and validates the whole table. `migrate.go` runs each migration in one transaction inside `main()`, so if any existing row anywhere holds a value not in the list, **every install fails to boot at once**. Having just learned the list was wrong for months, the odds of an unknown legacy value in real data are not negligible — and this is the one change whose failure mode is "nothing starts".
- **A migration per future status**, which couples every Go state-machine change to a DB change and, under this repo's rule that the migration lands first, makes each new status a two-PR sequence. On a state machine actively growing (#463 alone added three) that is continuous friction.
- **A rejected write becomes a 500**, not a degraded row.

### What I would do instead

A committed guard that diffs `model.go`'s const block against the `COMMENT ON COLUMN` text — mechanically what `reconcile-482.sh` does in this PR, including its refuse-on-empty behaviour — wired into `ci.yml`. It catches **both** failure modes this constraint is wanted for (the omission *and* the typo `'sheduled'`), it fails in CI rather than at 3am inside a halt, it takes no lock, it risks no boot, and it needs no migration per status.

That belongs in `scripts/` with its own test suite, following the `check-version-surfaces.sh` + `check-version-surfaces_test.sh` pattern. It is `devops-engineer`'s path for the script and `backend-architect`'s for anything Go-side — not this agent's, so it is not in this PR.

### If a CHECK is wanted anyway

Then: `NOT VALID` first (skips the full-table scan and the boot-time failure), in its own ordinal, **after** the guard above has run green for a release cycle so the list is known complete rather than assumed complete — and never in the same migration as a comment change.

mosamlife:
Correcting a claim I made on this PR: I said already-drifted rows could not be identified
retrospectively, because the halt reason is not stored on the run. **That is wrong**, and the
security review found the route.

`haltAgentRunWith` calls `recordRunHalted` (`agent_worker.go:691`), which writes an
`update.run.halted` audit event with `TargetType: "update_run"` and `TargetID: runID`,
gated on `ev.Changed` — which is `run.Status != RunHalted` (`agent_repo.go:322`), true in
exactly the drift scenario. `audit_log` is append-only for `wpmgr_app`
(`REVOKE UPDATE, DELETE, TRUNCATE`, `migrations/20260527130000_auth_multitenancy.sql:130`)
and carries no retention sweep.

So the drifted runs are enumerable:

```sql
SELECT r.tenant_id, r.id, a.created_at, a.metadata->>'reason'
FROM update_runs r
JOIN audit_log a
  ON a.target_type = 'update_run' AND a.target_id = r.id::text
WHERE a.action = 'update.run.halted' AND r.status = 'completed';
```

Post-fix that set is unreachable, so every row it returns is a historically drifted run — a
rollout an operator stopped that the system has been reporting as finished normally. No false
positives from the legitimate `completed -> halted` path, because that path ends on `halted`.

**This has not been run against production data.** It should be, before this issue closes —
operators may currently be looking at "completed" rollouts that were killed. The query is
read-only and answers the question in one pass.

The reasoning above was verified by reading the code, not by executing the query, and I am
recording that distinction rather than implying otherwise.

PR #492, opened 2026-08-19
fix(update): a scheduled run must not dispatch to a paused site (#463)

mosamlife:
Closes the last open decision on #463.

## The gap

The deferred dispatcher did not consider monitoring pause at all — `grep -ciE "paused" apps/api/internal/update/dispatch.go apps/api/internal/update/dispatch_repo.go` returned `0` for both against a control confirming `sites.monitoring_paused_at` exists. A run scheduled before a site was paused fired into it regardless.

## The decision: skip paused sites

This applies m117's existing policy rather than inventing new policy. The schema states pause "governs the SCHEDULE, never the operator", and must never stop "anything a person clicks". A deferred run **is on the schedule**: the click happened earlier, to put it there, and at fire time nobody is present. So pause governs it.

The practical case is the one that cannot be undone. People pause a site because something is wrong with it, and firing an update into a site somebody deliberately froze mid-incident is not recoverable. A skip is — they see it and re-run with one click.

## What changed

`apps/api/internal/update/dispatch_repo.go`

- Pause is read at **fire time**, inside the dispatch transaction, never at create time. A site paused *after* its run was scheduled is the whole scenario; a create-time read would miss exactly the case this exists for.
- **Per task, not per run.** A run spanning ten sites with two paused dispatches to eight and skips two. Killing the run because one site is frozen would be its own defect.
- The declined task becomes `skipped` with a detail naming the pause — not `failed`: nothing was attempted and nothing is broken.
- It is declined **before** `MarkScheduledUpdateTaskPending` and before any enqueue, so the task never takes an in-flight slot and no job exists that could reach the site.
- An unreadable pause state **fails closed**: the error rolls the claim back and a later tick retries, rather than dispatching into sites whose pause was simply not read.
- One set-based read, not an N+1, so a fleet-wide run does not hold the dispatch transaction open across hundreds of round trips.
- `tenant_id` is in the predicate and is load-bearing, not just defence in depth: this runs under `InAgentTx`, so RLS is not the boundary it is on a tenant-scoped path.

`apps/api/internal/update/dispatch.go` — a `paused_skipped` count (a **subset** of `skipped`, never an addend) on the dispatch audit entry and the log line.

## The run's terminal status when every site was paused: `completed`

Deliberate, and documented in the code so a later reader changes it on purpose. `schema.sql` defines `completed` as "All tasks reached a terminal state", which every task here reached — it has never meant "all work succeeded". It is not `failed` (nothing attempted, nothing broken), not `expired` (that is about the control plane missing its own window), not `halted` (reserved for an agent self-update wave with commands already out).

The decisive argument: an all-paused run is the **same shape** as the all-targets-busy run, which has always landed `completed`. Giving one of two identical situations a different status would report the same fact two ways. What keeps it honest is the per-task detail plus `paused_skipped`, rather than a new run status — which would be a wire-contract change across `schema.sql`, the OpenAPI enum and the dashboard to say what the task rows already say precisely.

## Proof

`apps/api/tests/gh463_paused_site_dispatch_integration_test.go`, run as `wpmgr_app` through `update.NewRepo` and `site.NewRepo` — never superuser, with an executable role guard asserting `NOT rolsuper AND NOT rolbypassrls`. Pause is applied through `site.Repo.PauseMonitoring`, the same call the operator's button makes, not by writing the column by hand.

"No command reaches the paused site" is asserted as the **absence of that task's enqueued job**, not as a status: `update.Worker.Work` is reachable only through a `TaskArgs` job.

RED, with the real failure planted (`monitoring_paused_at IS NOT NULL` → `false`, i.e. pause never consulted):

```
--- FAIL: TestGH463_PausedSiteIsSkippedAndTheUnpausedSiteStillDispatches (1.61s)
    a job was enqueued for the PAUSED site: the update will reach a site the operator deliberately froze
    enqueued 2 jobs, want exactly 1 (the active site only)
    outcome = {Claimed:true Dispatched:2 Skipped:0 PausedSkipped:0 Status:running}
    paused site's task = "pending", want "skipped"
--- PASS: TestGH463_ResumingBeforeTheRunFiresDispatchesNormally (1.36s)
--- FAIL: TestGH463_ARunWhoseSitesAreAllPausedCompletesWithoutContactingAnySite (1.36s)
--- PASS: TestGH463_PausedSkipRanAsTheApplicationRole (1.32s)
```

GREEN, restored — `go test ./tests/ -run 'TestGH463' -count=1 -v` passed all 20 GH463 tests in 29.053s, the 15 pre-existing Phase 0/Phase 1 tests included.

That mix is the point in both directions. The two pause tests are red without the change and green with it. `ResumingBeforeTheRunFiresDispatchesNormally` passes **either way** on purpose — it is the control that proves the change does not block honest work, and the Phase 1 suite (`ABusyTargetIsSkippedAndTheRunSurvives` especially, the neighbouring skip path) is the over-fire guard.

## Notes for other layers

- **Frontend.** The reason reaches the client — `update_tasks.detail` is mapped at `apps/api/internal/update/handler.go:589` and `use-row-update.ts` exposes it as `progress` — but the `skipped` badge in `apps/web/src/features/updates/available-updates-card.tsx:577` renders the bare word "Skipped" with no detail, unlike `rolled_back` which appends the reason. A paused-site skip will currently read as unexplained. `apps/web/**` is out of this PR's scope.
- No schema, migration, `db/query/**` or OpenAPI change. Nothing to regenerate.

## Summary by CodeRabbit

- **New Features**
  - Scheduled monitoring runs now check site pause status when dispatch begins.
  - Tasks for paused sites are skipped without contacting those sites, while unpaused sites continue normally.
  - Paused-site skips are clearly identified in run details and audit records.
  - Runs containing only paused sites now complete successfully without dispatching requests.

- **Bug Fixes**
  - Resuming a site before a scheduled run fires allows it to be processed normally.


mosamlife:
Integration suite (`make test-integration`, not run by `ci.yml`):

- **Run 1: FAILED.** `FAIL github.com/mosamlife/wpmgr/apps/api/tests 747.312s`. I piped that run to `tail -30` and so **did not capture the failing test name** — my error.
- **Run 2: PASSED.** `make test-integration > /tmp/gh463-integration.log 2>&1` → `EXIT=0`, `ok github.com/mosamlife/wpmgr/apps/api/tests 719.547s`, `ok .../tests/contract 1.221s`.
- Targeted follow-up on the timing-sensitive candidates suggested by run 1's trailing log lines (`vuln: feed rate limited; skipping this cycle`, `<31m since last Wordfence request`): `go test ./tests/ -run 'TestFeedWorker' -count=3` → `ok 51.239s`. Did not reproduce.

**Stating this plainly rather than calling it green:** the first failure is unattributed. What I can say is that it did not recur across a full clean suite run plus three passes of the most plausible candidate, and that this diff is confined to `internal/update`'s dispatcher, whose own coverage — all 20 `TestGH463*` tests, the 15 pre-existing Phase 0/Phase 1 ones included — passed in every run. Worth a second suite run on CI or by a reviewer before merge rather than taking run 2 as settling it.

mosamlife:
Confirmed real, and deliberately not changed in this PR. Recording why.

The read is on the caller's transaction, so under READ COMMITTED a pause that
commits after the `SELECT` and before the dispatch commit is invisible to it.
The window is the dispatch transaction's own duration. `git grep` for
`monitoring_paused_at` across `internal/update/` returns matches in
`dispatch_repo.go` only — the apply worker does not re-check, so the report is
accurate on that point too.

Two reasons it stays:

**Pause has never gated updates at all.** This PR adds the first such gate, and
only for the deferred path, because a scheduled run fires unattended at a time
the operator is not present — they paused in between and reasonably expect
nothing to reach the site. An immediate update is the operator asking right now,
and is intentionally not filtered. So the behaviour under the race is exactly
the pre-existing behaviour, not a new hole.

**Closing it costs more than it buys.** True serialization needs the dispatch to
lock the `sites` rows it is about to read, which puts the operator's pause button
in contention with the dispatcher. Trading an interactive control's latency
against a millisecond window is the wrong side of that deal.

The cheaper mitigation — re-checking at apply time, in the worker that actually
contacts the site — is plausible but changes behaviour for immediate updates too
unless it is gated on the run being a scheduled one. That is a real design
question rather than a fix, and it is not this PR's scope.

mosamlife:
Confirmed real, and deliberately not changed in this PR. Recording why.

The read is on the caller's transaction, so under READ COMMITTED a pause that
commits after the `SELECT` and before the dispatch commit is invisible to it.
The window is the dispatch transaction's own duration. `git grep` for
`monitoring_paused_at` across `internal/update/` returns matches in
`dispatch_repo.go` only — the apply worker does not re-check, so the report is
accurate on that point too.

Two reasons it stays:

**Pause has never gated updates at all.** This PR adds the first such gate, and
only for the deferred path, because a scheduled run fires unattended at a time
the operator is not present — they paused in between and reasonably expect
nothing to reach the site. An immediate update is the operator asking right now,
and is intentionally not filtered. So the behaviour under the race is exactly
the pre-existing behaviour, not a new hole.

**Closing it costs more than it buys.** True serialization needs the dispatch to
lock the `sites` rows it is about to read, which puts the operator's pause button
in contention with the dispatcher. Trading an interactive control's latency
against a millisecond window is the wrong side of that deal.

The cheaper mitigation — re-checking at apply time, in the worker that actually
contacts the site — is plausible but changes behaviour for immediate updates too
unless it is gated on the run being a scheduled one. That is a real design
question rather than a fix, and it is not this PR's scope.

PR #519, opened 2026-08-23
docs(adr): amend ADR-061 (A1-A13), close ADR-062's principal blocker, add ADR-063

mosamlife:
Records an architecture decision that has already been made and evidenced. Documentation only; no code paths touched.

## ADR-061 — amended in place, Status stays Accepted

No decision text was rewritten. A new `## Amendments (2026-08-23)` section carries eight items, and each amended decision carries a pointer to it where a reader will hit it.

| # | Change |
|---|---|
| A1 | Clarifies Decision 1 without amending it. The fan-out argument is scoped to fleet-aggregate questions; under Decision 2 a per-site tool call writes a proposal row, so it costs one INSERT and zero round trips, and dispatch happens later on a worker after a human approves. The per-site clause in Consequences is activated as the sanctioned route. The credential argument is untouched and remains load-bearing. |
| A2 | Records the agent's existing signed-command surface, with the two commands that measure it. Corrects a reading, not a decision: what is missing for per-site work is model-facing exposure, not capability. Decision 7's content-write premise is re-verified and remains true. |
| A3 | Amends Decision 6. The revisit threshold is measured against `tools/list` size specifically and re-measured per release, replacing a round number. Records the three-layer registry. Retires the client-approval-UI argument, with the reasoning: Decision 2 removed that UI from the trust model, so the argument binds a different architecture. The tail-is-not-a-registry half is retained and confines Layer C. |
| A4 | Decides concurrency: a per-organization ceiling (default 4) on the assistant dispatch queue plus a per-site exclusive advisory lock, both following patterns already in `internal/update` and `internal/org`. Fixes the **unit** as the per-site step, records that runs are not capped at all, and gives the conversion. Records that the withdrawn wireframe figure was mock copy with no decision behind it. |
| A5 | Bounds the no-automation-may-approve foreclosure against the scheduler: a scheduler that only proposes is permitted, nothing scheduled may approve. |
| A6 | Names the protocol. MCP, **target 2025-11-25, floor 2025-03-26**, refuse below the floor with the reason surfaced, transport, and resources/prompts scoped out of Phase 1. The ADR previously never said MCP while the shipped surfaces already published an `/mcp` endpoint. |
| A7 | Marks Decision 7 superseded by ADR-062 and retires its Consequences bullet. |
| A8 | Adds one line under *Two things that were expensive to learn*: prompt-level safety is not a boundary now has external corroboration. |

The signing statement ADR-061 already corrects is not reintroduced — signing gives forgery and tampering resistance while the key is trusted, and nothing in A2 or A4 claims more.

## ADR-062 — new, Status Proposed

`docs/adr/ADR-062-assistant-surface-phase-2-content-operations.md`. **Proposed, not Accepted — it blocks Phase 2 code.** Carries the nine-item acceptance checklist and records why Decision 7's market conclusion failed, why byte comparison was the wrong verify test and what replaces it, that the unit of work is fleet content operations rather than single-site authoring, snapshot-before-destructive-write and verify-and-rollback as platform primitives rather than per-integration ones, capability detection over the signed channel with refusals naming the requirement and the site's actual version, dynamic-data binding as the write-reduction strategy, field writes through the field plugin's own API rather than raw meta, and the block-editor fork with both branches costed.

Item 2 — **which WordPress principal a content write executes as** — is open and recorded as a hard blocker with three candidate answers and their costs. It routes to `security-reviewer` before any content code.

## Second commit — two under-specifications closed

An adversarial cross-artifact consistency check found two places where published artifacts disagreed with each other and the ADR arbitrated neither. Both were the ADR being under-specified rather than wrong, so both fixes are additions.

**A6 had a target version and no floor.** One artifact accepted a pre-floor protocol revision in its negotiation window; another drew a refusal screen for that same revision and gave a reason tied to the approval flow. The refusal wins, because its reason is tied to the product: below the floor the protocol drops fields the approval flow depends on, and this ADR exists to make a named human's approval checkable. A6 now records **three** values — target `2025-11-25`, floor `2025-03-26`, and refuse-below-the-floor with the reason surfaced to the operator rather than a bare version mismatch. The target version is now corroborated against the WordPress project's published MCP schema package rather than asserted.

**A4 stated its ceiling in steps while artifacts rendered it as run slots and as operations.** Three nouns for one limit is how the withdrawn ceiling figure survived. A4 now fixes the unit as the **per-site step**, states that runs are not capped at all, and gives the conversion: a run over N sites contributes at most `min(N, ceiling)` concurrently dispatching steps and never more than one step per site, because the per-site lock serializes a site's own steps whatever the ceiling allows. Counting the ceiling in runs would understate fleet load by roughly the size of the fleet.

### ADR-062 — the principal is decided, and the ADR stays Proposed

**Item 2 is closed.** The WordPress execution principal is a dedicated service user with a custom role, created by the agent: no password, no login path, no session, and deliberately **no `unfiltered_html`**, so `wp_kses_post` runs and a content write cannot inject script.

The cost is stated rather than argued away. It is a privileged object on every managed site, and the ADR does not let "not really privileged" stand as an assertion — it specifies an **attack-test suite as a ship gate**, not a review item: authenticate by password, mint an application password, reach it through the autologin path, establish a REST or admin session, and enumerate its capabilities. Each is planted as a *passing* attack first and watched go red before it is watched go green. Two other costs are on the record: the site's own admin can delete it, which must surface as a typed `principal_missing` refusal with a recreate remedy and never a silent fallback to a different principal; and it is a new visible row on the customer's Users screen.

What it forecloses is recorded as deliberate: **making an AI change look human-authored** (attribution is legible by construction; a future per-site `post_author` override moves the display field only, never the executing principal), and **arbitrary embed or HTML writes** (without `unfiltered_html` a prompt-injection payload reaching the content path cannot become executable markup).

**The checklist now marks decided against open**, and the ADR carries the command that counts both — refusing on zero, so a reformatted list cannot print `0 open` and read as an ADR ready to accept.

**The wave plan closes item 7.** Wave 1 has **no page builder and no Gutenberg writes**. Wave 2 gets **exactly one builder, Elementor**. Wave 3 is Gutenberg static blocks and Bricks, gated on the fleet query. Everything else is out of scope with a one-line reason each. Two honesty statements travel with the plan wherever it is rendered: **the ordering is reasoned, not measured**, and **the fleet-inventory query that would measure it is a pre-development task nobody has run**. The criteria do not change if the query surprises us; the ordering does.

**ADR-062 stays Proposed**, and the ADR now carries a `Why this is still Proposed` section rather than leaving it to inference: accepting it lifts its own gate on content-write code, ADR-060's precedence puts it behind open rank-1 work, two checklist items need a measurement and a mechanism rather than an argument, the wave plan is provisional on a query nobody has run, and `security-reviewer` has not seen the principal or the four-command token scoping.

### ADR-061 — amendments A9 through A13

| # | Change |
|---|---|
| A9 | **The most consequential sequencing fact in the plan.** ADR-060's freeze clause fixes ordering *inside* Phase 1: the MCP endpoint is a new externally-reachable surface, and the site-scope chokepoint and the audit fail-open are open auth-boundary items, so they close **before** the endpoint ships. Not "should" — the clause is absolute and it is ADR-060's only absolute. |
| A10 | Audit is **fail-closed for this surface**, in one sentence, noting the existing write helper is documented fail-open and that opting in is deliberate. Reads are included; the chain lock becomes a throughput ceiling that must be measured rather than discovered. |
| A11 | **Site scoping is in v1** as an application-layer chokepoint, with the six pieces of work listed. Reaffirms that no document may imply the database enforces it — the scope and allowlist columns are stored and enforced by nothing today. |
| A12 | **Protocol**: the three cases the code must not conflate. An absent `MCP-Protocol-Version` header means assume the floor `2025-03-26`, **not** a 400; an unsupported or unparseable value MUST get a 400; the negotiated version governs. No AI client's public documentation mentions the header, so the server logs what it actually receives from day one. |
| A13 | **Skill and prompt content is data, never permission.** Authorization resolves outside the model loop, so instruction text asking to skip approval is denied identically to text that does not. The fencing rule is Phase 1 even though the store itself is later: the fence must predate the first untrusted string, not the first skill. |

### ADR-063 — new, Status Accepted

`docs/adr/ADR-063-licensing-and-third-party-reuse.md`. Records repository licensing as it actually is, verified by commands the file carries rather than quoted from memory: the root `LICENSE` is **AGPL-3.0** and covers `apps/api`; `LICENSE-AGENT` is **MIT** and covers `apps/agent` and `apps/tracker`.

The consequence is an asymmetry that points the opposite way from the one people expect: **GPL-2.0-or-later components cannot enter the MIT plugin without relicensing it**, while the AGPL control plane has no such conflict, because the repository is already public and already AGPL. **This independently supports siting the MCP surface in the control plane** — a second argument for ADR-061 Decision 1 that does not depend on the credential one.

Adds the four reuse dispositions for engineers (adopt upstream, interface-compatible only, reimplement from behaviour, do not use), and records that a protocol, a published specification and a public API shape are not copyrightable expression the way source code is. Two items are left **open for an owner ruling**: the plugin header declares `License: MIT` while `readme.txt` declares `GPLv2 or later`, and the decision to keep MCP out of the plugin for licence reasons.

`docs/adr/README.md` is kept in step: ADR-063 registered, next free number advanced, amended statuses transcribed.

## Gates run

- Docs vocabulary check, copied out of `.github/workflows/ci.yml` in this pass and run over `docs/**.md` and root markdown: no output, all three commits. **Proved it fires** rather than trusting a silent pass: planted a banned name in a scratch file under `docs/`, watched it go red naming the file, removed it, watched it go green.
- `make hooks-status`: pre-push installed and current.
- `make check-versions-test` then `make check-versions`: pass, self-test first.
- `node apps/marketing/scripts/check-copy.mjs`: pass.
- Every count in either document is the output of a command run while writing it, with the command beside it. Neither ADR hard-codes a checklist count; ADR-062 carries the command that prints it.

Draft. Please do not merge before review.

### DIFF
--- a/docs/adr/ADR-061-assistant-surface-phase-1.md
+++ b/docs/adr/ADR-061-assistant-surface-phase-1.md
@@ -1,7 +1,7 @@
 # ADR-061 — Assistant surface, Phase 1: control-plane server and out-of-band approval
 
-**Status:** Accepted (amended 2026-08-23) · **Date:** 2026-08-22
-**Supersedes/relates:** ADR-060 (this work sits in its "constrained automation" phase, ahead of differentiation), ADR-057 (unaffected; the capability model here is a separate axis from per-site security policy), ADR-062 (Proposed — supersedes Decision 7 below, per amendment A7).
+**Status:** Accepted (amended 2026-08-23 and 2026-08-24) · **Date:** 2026-08-22
+**Supersedes/relates:** ADR-060 (this work sits in its "constrained automation" phase, ahead of differentiation — and its freeze clause fixes the ordering *inside* this phase, per amendment A9), ADR-057 (unaffected; the capability model here is a separate axis from per-site security policy), ADR-062 (Proposed — supersedes Decision 7 below, per amendment A7), ADR-063 (licensing and third-party reuse — its conclusion independently supports Decision 1's siting).
 
 This ADR records the locked Phase 1 design for the assistant surface: an
 endpoint on the control plane that lets an AI assistant read a fleet and
@@ -18,6 +18,15 @@ on Decisions 1, 6 or 7.** The status stays Accepted and no decision text below
 was rewritten. Eight amendments are recorded at the end of this file: Decision 1
 is clarified rather than changed, Decision 6 is amended, Decision 7 is
 superseded by ADR-062, and four items ADR-061 was silent on are decided.
+
+**Amended again 2026-08-24 — read [Amendments](#amendments-2026-08-24) before
+planning any Phase 1 work.** Five further amendments, A9 through A13. **A9 is the
+one to read first**: ADR-060's freeze clause fixes the ordering *inside* Phase 1,
+so the gate work closes before the endpoint ships. The rest make the audit posture
+fail-closed for this surface, put site scoping in v1 as named work, settle three
+protocol-header cases the code must not conflate, and state the rule that skill
+and prompt content is data and never permission. No decision text below is
+rewritten by any of them.
 
 ---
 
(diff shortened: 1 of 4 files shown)


mosamlife:
## Review against `main` @ `ab20cc63` — four things to fix before this lands

I re-verified every code citation, line number, count and command in this diff against current `main`. **All of them still hold**, including the 68-command / 55-name counts, `update/worker.go:60`, `rucss/worker/worker.go:405`, the audit fail-open doc comment, the m120 migration header, and both external claims. The PR is 35 commits behind but nothing in it has gone stale. The habit of carrying the verifying command next to the figure is what made that checkable in minutes.

The problems are not factual drift. They are four internal conflicts.

### 1. A3 says Decision 6 is unchanged while changing it

Decision 6 reads: *"A small flat set of fleet-level tools. No meta-tool facade, no per-site command mirror."* A3 states the text below is unchanged, then adds **Layer B** (per-site proposal tools) and **Layer C** (a dispatcher) — the two things that sentence forbids by name.

Retiring one of Decision 6's three arguments is not the same as amending the decision. A reader who follows the ADR's own instruction concludes per-site tools are still forbidden; a reader who reads A3 concludes they are Layer B.

**Fix:** frame A3 as an explicit scope amendment to Decision 6, or split Layers B and C into their own ADR. Also reconcile Layer C's generated-enum argument with Decision 5, which says a capability is handled *by absence from the registry* rather than by a gate inside it — a runtime-computed argument constraint is a gate inside the dispatcher.

### 2. ADR-062 cites a fail-open mechanism as its snapshot precedent

ADR-062 requires that every destructive content write is preceded by a snapshot, and cites the agent's file-write staging as proof this is "not a new capability."

That mechanism is best-effort and documented as such: `apps/agent/includes/commands/class-file-write-command.php:257-258` says *"Errors are silenced — failure here does NOT block the write."* `stageBackup()` returns silently on four separate paths and the caller ignores its return value before proceeding to the write.

So it **is** a new capability in the one respect that matters: the existing one is allowed to fail open, and a mandatory one is not. Checklist item 4 is correctly open, but for a different reason, so it could be closed while the rule remains false.

**Fix:** either a second sub-item for making the extended mechanism fail-closed, or state plainly in ADR-062 that the precedent is fail-open and the new one will not be.

### 3. Checklist item 9 cannot ever close

Item 9 reads *"this item closes when this ADR is Accepted"*, while the ADR states *"every `[ ]` is a reason this ADR is still Proposed"* and *"Every checklist item closed"* is the precondition for Phase 2 work.

Item 9 closes only on acceptance; acceptance is blocked by every open item including item 9. The self-counting checklist — the most rigorous-looking device in the ADR — will report a permanent floor of one open item forever.

**Fix:** mark it `[x]` (A7 already did the work) or strike it from the acceptance checklist. It is a bookkeeping consequence of acceptance, not a gate on it.

### 4. ADR-063's standing rule 4 is breached by ADR-062 in the same commit

Rule 4 forbids naming a competitor product in any committed document, carving out only `apps/marketing/**`. ADR-062 — a committed document added in this same PR — names nine page builders, several as headings.

CI passes, correctly: the enforced ban list at `.github/workflows/ci.yml:365` is narrower, and its own preamble at `:360-362` exempts *"functional source references (conflict-detection slugs, plugin-signature seeds, compat handling)."*

So the defect is in ADR-063, not in ADR-062 or in CI. The rule as written is stricter than the gate that enforces it and omits the exemption CI documents inline.

**Fix:** give rule 4 the integration-target exemption explicitly. A newly accepted rule that its sibling breaches in the same commit is the kind that gets informally suspended and never recorded.

### One thing the amendments understate

A9 states *"the chokepoint does not exist yet"* and A11 reads as a six-item build from zero. But `authz.RequireSiteAccess` (`apps/api/internal/authz/middleware.go:26-65`) already enforces `AllowedSiteIDs` on `:siteId` routes, and `apikey.go:298` says so explicitly. It has the known fail-open zero-value shape that Decision 4 itself names (`middleware.go:38-41`).

**Harden the existing chokepoint rather than building a second one — two boundaries are worse than one.** A11 item 3 in particular ("empty allowlist resolves to zero sites") is exactly the defect in that predicate's zero-value path, and reads much stronger as *fixing the existing function* than as a property of a new one.

---

Not merging and not changing ADR status — this is review only. A9 is the strongest thing in the PR, and its reading that the freeze clause binds *inside* Phase 1 is one I could not fault against ADR-060's text. Worth noting separately that the freeze clause is referenced by zero enforcement artifacts today: no CI step, no PR template, no rule file. It is a standing check that nothing checks.

mosamlife:
Closing this draft. It carried three things; here is where each went.

- **ADR-062** (Assistant surface, Phase 2: governed fleet content operations) → merged as #549.
- **ADR-063** (Repository licensing, and the rules for third-party reuse) → merged as #554.
- **ADR-061 amendments A1-A13** — the only part still stuck here — extracted to their own PR against current `main`: #563.

#563 also fixes a framing issue in A3 that this draft had: it claimed Decision 6's text was "unchanged" while introducing a per-site dispatcher (Layer C), which is exactly the meta-tool facade Decision 6 forbids by name, and left that dispatcher's runtime-gated argument enum unreconciled with Decision 5's absence-from-registry model. #563 drops Layer C (recorded as considered and rejected, not carried forward) and reframes the curated per-site tool set (Layer B) as already operating within Decision 6 as written rather than amending it. Details are in #563's description.

Branch kept, not deleted.

mosamlife:
Superseded by #549 (ADR-062), #554 (ADR-063), and #563 (ADR-061 amendments A1-A13). See comment above for the breakdown.

Issue #553, opened 2026-08-27
Site inventory has no age, and as_of reports the heartbeat instead


mosamlife:
Found during the fleet-inventory analysis for the assistant roadmap (slice S2).

## The claim we make

`GET /sites/{id}/updates` returns an `as_of` timestamp. `apps/api/internal/site/handler.go:577-579` stamps it from `s.UpdatedAt`:

```go
if !s.UpdatedAt.IsZero() {
    out.AsOf = gen.NewOptNilDateTime(s.UpdatedAt)
}
```

## Why it is wrong

`sites.updated_at` is bumped by the **heartbeat**, every 60 seconds, by `TouchSiteHeartbeat` (`apps/api/db/query/site_connection.sql:167-172`):

```sql
UPDATE sites
SET last_seen_at      = now(),
    missed_heartbeats = 0,
    updated_at        = now()
```

That statement never touches `components`, which is where the plugin and theme inventory actually lives (`apps/api/db/schema.sql:309`).

So `as_of` answers "when did this site last say hello", not "when was this inventory collected". The two are unrelated, and the error is **one-directional**: it can only ever overstate freshness, never understate it.

**Failure scenario.** A site's WP-Cron dies. The agent's 30-minute metadata refresh (`apps/agent/includes/class-scheduler.php:179-182`) stops running, so `components` freezes. The heartbeat keeps working, so `updated_at` keeps advancing. The dashboard shows an inventory timestamp seconds old while serving plugin data that is arbitrarily stale — potentially months. Nothing on the control plane detects this, because there is **no control-plane-side scheduled inventory refresh at all**; the refresh is entirely agent-driven.

The existing sweeper does not cover it either: it degrades at 300s and disconnects at 900s on **agent reachability**, which is exactly the thing that is still healthy in this scenario.

## Why it matters beyond the timestamp

A stale inventory is not a cosmetic problem. It drives the update list an operator acts on, and it is the only source we would have for any fleet-wide question of the form "how many of my sites run X, at version Y". A number derived from it today carries no defensible as-of, because the column that would date it does not exist:

```
$ grep -n components apps/api/db/schema.sql
309:    components  jsonb  NOT NULL DEFAULT '{}'::jsonb,
```

No `components_updated_at`, and `components` is never cleared or flagged when a site goes `disconnected`.

## Suggested shape

1. Add a dedicated inventory timestamp column, set only when the inventory is actually written — not by the heartbeat.
[...]

Issue #509, opened 2026-08-21
NOT NULL DEFAULT 0 on measurement columns manufactures data: 'never measured' is stored as a real zero


mosamlife:
## What is wrong

`apps/api/db/schema.sql:4111,4113`:

```sql
oc_latency_ms           integer     NOT NULL DEFAULT 0,
oc_used_memory_bytes    bigint      NOT NULL DEFAULT 0,
```

`NOT NULL DEFAULT 0` means **"no heartbeat has arrived yet" is stored as a real `0`**, indistinguishable from a genuine zero reading. The object-cache panel renders `0.0 ms` before a site has ever reported — a measurement that was never taken, displayed as a measurement.

The nullable sibling three lines down shows the contrast:

```sql
oc_hit_ratio_pct        numeric(5,2),      -- line 4114, nullable
```

That one renders as an absence, correctly, because it can actually be absent.

## Why this is the GH #460 defect, one layer down

#460 was a nullable uptime figure typed as a bare `float64`: nil became `0`, and a never-probed site rendered as 90 days of solid outage. The report was worse than showing nothing because it looked like data.

Here the same substitution happens **in the schema** rather than in the display layer. By the time the value reaches Go it is already a `0` with no way to tell it apart from a real reading, so no amount of care in the API or the dashboard can recover the distinction. Every layer above is doing the right thing with a value that was falsified before they saw it.

This is worth naming as a class: `NOT NULL DEFAULT 0` on a **measurement** column is almost always wrong. A default is appropriate for a setting, a counter, or a flag. For an observation, it manufactures data.

## How it was found

During GH #508's inverse audit. `ObjectCachePanel.tsx:1361-1362` coalesced both fields with `?? 0`, which looked like the #460 shape at the display layer. Investigation showed the opposite: the fields genuinely cannot be null on the wire, so the coalesce was dead code and was removed. Three layers agree they are always present:

- `apps/api/db/schema.sql:4111,4113` — `NOT NULL DEFAULT 0`
- sqlc reads them as `int32`/`int64`, not pointers
- `apps/api/internal/objectcache/dto.go:55,57` — value types with no `omitempty`, so `encoding/json` always emits a number

So the declaration was right and the defensive code was wrong — but only because the schema had already destroyed the information.

## Done when

- The two columns can express "not yet measured" distinctly from a real zero.
[...]

PR #559, opened 2026-08-27
feat(db): m121 - a dedicated age for the site inventory (GH #553)

mosamlife:
Closes the database half of #553. **Schema only — no reader changes.**

## The defect

`GET /sites/{id}/updates` stamps `as_of` from `sites.updated_at`. That column is bumped every 60 seconds by `TouchSiteHeartbeat` (`apps/api/db/query/site_connection.sql`), which never touches `components`:

```sql
UPDATE sites
SET last_seen_at      = now(),
    missed_heartbeats = 0,
    updated_at        = now()
```

So `as_of` answers "when did this site last say hello", not "when was this inventory recorded". The error runs **one way only**: `updated_at` is always `>=` any instant `components` was written, so freshness can be overstated and never understated. A site whose WP-Cron dies stops refreshing its inventory while its heartbeat keeps advancing `updated_at` — the dashboard then shows a seconds-old timestamp over plugin data that may be months stale.

## The change

One nullable column on `sites`:

```sql
ALTER TABLE "public"."sites"
    ADD COLUMN IF NOT EXISTS "components_updated_at" timestamptz NULL;
```

Written on exactly one statement, `UpdateSiteMetadata` in `apps/api/db/query/sites.sql` — the only statement in the tree that writes `components`.

## Existing rows get NULL. No default, no backfill.

`NULL` means "we have never recorded when this was collected", which is the truth for every row that exists when this applies. The three tempting backfills all manufacture an observation that was never taken:

- **`updated_at`** is the defect itself — it would permanently store the heartbeat instant this column exists to stop reporting, indistinguishable from a genuine stamp.
- **`last_seen_at`** is the same substitution with the same one-way error.
- **`now()`** would claim the entire fleet was collected at boot — the most flattering and most misleading value available.

This is #509's defect one table over: `NOT NULL DEFAULT 0` on a measurement column made "never measured" indistinguishable from a real zero, and no care in the API or the UI can recover a distinction the schema already destroyed. A default is right for a setting, a counter or a flag, and wrong for an observation.

## Naming: `components_updated_at`, not `components_collected_at`

The value is `now()` on the **control plane** at the moment it persisted the document, not the instant the agent walked the plugin list on the WordPress host. The agent's metadata payload carries no collection timestamp, so the collection instant is not knowable here. Calling it `collected_at` would assert a fact about the remote host we do not have — the same manufacture as a backfill, just with a smaller error bar. `host_provider_checked_at` (m28) is the house precedent for a when-did-we-last-look field.

It is written on **every** metadata push, even when the document is byte-identical. Freshness is about when the claim was last confirmed, not when it last changed; a write-only-if-different rule would freeze the stamp on a stable site and reinvent the bug for exactly the fleet that is behaving correctly.

## Disconnected sites: left alone, deliberately

No connection transition touches this column. The stamp goes stale and **the staleness is the signal** — any reader computes `now() - components_updated_at` and gets the truth regardless of `connection_state`. The decision and its alternatives are recorded in the migration comment, because the column outlives the issue:

- **Nulling it on disconnect** would erase a fact that really happened, and collapse "recorded three months ago" into "never recorded" — the same conflation running backwards. Worse, nothing clears `components` itself on disconnect, so the result would be *undated inventory data*: strictly worse than dated stale data, because stale-and-dated can be reasoned about.
- **A separate stale flag** would be a second column that can disagree with the first, for a question the first already answers exactly. The threshold ("how old is too old?") is policy that belongs where it can be tuned per caller, not frozen into a boolean at write time.

Re-enrollment (`AttachAgentToSite`) does not touch it either, for the same reason: that statement does not rewrite `components`, so the surviving inventory keeps its true, older stamp. The column and the data it dates always move as a pair.

## What is NOT in this PR

- **`apps/api/internal/site/handler.go` is unchanged.** Stamping `as_of` from the new column is `backend-architect`'s slice and lands after this one — a change spanning a migration and Go code is two agents in sequence. The API side has not been forgotten; it is deliberately sequenced second, and until it lands `as_of` still reports the heartbeat.
- **No control-plane staleness sweep.** #553 item 4 is a real gap and is deliberately left open — it needs a sweep, a threshold and an alert route. This migration is the column that makes such a sweep possible; it is not the sweep.
- **No index.** The only reader that exists after this lands finds its row by primary key. `sites` is the hottest-written table in this schema, so a speculative index is paid on every heartbeat, sweep, probe and metadata push. The migration records what would justify one later and says to measure with `EXPLAIN ANALYZE` as `wpmgr_app` first.

## Write paths

`UpdateSiteMetadata` is the only one, established by grep rather than assumption:

- every `db/query/*.sql` mentioning `components` → only `sites.sql`
- every raw `UPDATE sites` / `INSERT INTO sites` in non-test Go → two writers, `diagnostics.UpdateSiteTimezone` and `diagnostics.SetSiteHostProvider`, neither of which touches `components`
- caller chain: agent metadata push → `Service.ApplyAgentMetadata` → `Service.ApplyMetadata` → `pgRepo.UpdateMetadata` → `sqlc.UpdateSiteMetadata`

The heartbeat path (`Heartbeat` / `RecordHeartbeat`) never reaches it. Both the query file and the migration carry a note that `TouchSiteHeartbeat` must never add this column.

## RLS

Nothing added, and this was read rather than assumed. Row-level security is row level, so a new column on an existing table inherits the table's policies with no new policy and no new grant. `sites` already carries `ENABLE` + `FORCE ROW LEVEL SECURITY`, `sites_tenant_isolation`, `sites_enroll`, `sites_agent`, the **RESTRICTIVE** `sites_site_scope` (m19), `sites_shared_read` (m22) and `sites_client_read` (m66). `sites_site_scope` is `FOR ALL`, so it constrains the metadata-push UPDATE as well as reads. No table is created, so the site-keyed-table checklist does not apply — `sites` *is* the site table.

Grants on `sites` are table-level, which covers columns added later; there are no column-level grants on this table.

## Converge path

**None required**, and this is a positive statement rather than an omission. The ordinal `20260823000000` is new and unique — the highest on any ref at the time of writing was m120's `20260822000000`. It corrects no earlier migration and edits no applied file, so there is no database on an earlier version of it to converge. Fully idempotent (`ADD COLUMN IF NOT EXISTS`), nothing dropped, no existing row written, no backfill to get wrong.

## Summary by CodeRabbit

* **New Features**
  * Site records now track when component metadata was last updated.
  * This timestamp is included in site details and listing responses.

* **Bug Fixes**
  * Component metadata updates now consistently refresh the associated update time, even when the content itself does not change.
  * Heartbeat updates remain separate and no longer affect this timestamp.

### DIFF
--- a/apps/api/db/query/sites.sql
+++ b/apps/api/db/query/sites.sql
@@ -197,6 +197,23 @@ RETURNING *;
 -- name: UpdateSiteMetadata :one
 -- Tenant-scoped metadata update (used by the agent path inside the resolved
 -- site's own tenant scope).
+--
+-- m121 (GH #553): this is the ONLY statement in the tree that writes
+-- sites.components, so it is the only one that may write
+-- components_updated_at. Verified by grepping every db/query/*.sql and every
+-- raw UPDATE/INSERT on sites in non-test Go; the other two raw writers
+-- (diagnostics.UpdateSiteTimezone, diagnostics.SetSiteHostProvider) touch
+-- neither column. The two must move together: a write of components without
+-- its stamp re-creates the undated-inventory bug, and a write of the stamp
+-- without components would date an inventory that was not refreshed.
+--
+-- It is now() on the CONTROL PLANE, not the agent's collection instant, and it
+-- is written on every push even when the document is unchanged -- "as of T this
+-- was the inventory" is a truer statement than the previous stamp regardless.
+--
+-- NOTE FOR THE HEARTBEAT: TouchSiteHeartbeat (db/query/site_connection.sql)
+-- must NEVER add this column. Bumping updated_at on a 60s heartbeat that does
+-- not touch components is the whole of GH #553.
 UPDATE sites
 SET wp_version   = $3,
     php_version  = $4,
@@ -205,6 +222,7 @@ SET wp_version   = $3,
     active_theme = $7,
     agent_version = $8,
     components   = $9,
+    components_updated_at = now(),
     last_seen_at = now(),
     health_status = 'healthy',
     updated_at   = now()
--- a/apps/api/db/schema.sql
+++ b/apps/api/db/schema.sql
@@ -307,6 +307,35 @@ CREATE TABLE sites (
     -- components holds the installed plugins/themes inventory as JSONB (M2): a
     -- normalized child table can come later; JSONB is sufficient for M2.
     components  jsonb       NOT NULL DEFAULT '{}'::jsonb,
+    -- m121 (GH #553): when the control plane last WROTE components, whether or
+    -- not the document changed. The inventory's own age, which updated_at is
+    -- not: TouchSiteHeartbeat bumps updated_at every 60s and never touches
+    -- components, so as_of stamped from updated_at can only ever OVERSTATE
+    -- freshness -- a site whose WP-Cron died keeps heartbeating while its
+    -- inventory freezes for months.
+    --
+    -- NULLABLE WITH NO DEFAULT, ON PURPOSE. NULL means "we have never recorded
+    -- this", which is the truth for every row predating m121, and there is no
+    -- backfill. Backfilling from updated_at or last_seen_at would write the
+    -- heartbeat instant this column exists to stop reporting; backfilling from
+    -- now() would claim the whole fleet was collected at boot. All three
+    -- manufacture an observation that was never taken -- GH #509's defect, where
+    -- NOT NULL DEFAULT 0 on a measurement column made "never measured"
+    -- indistinguishable from a real zero. A default is right for a setting, a
+    -- counter or a flag; this is an observation.
+    --
+    -- It is the CONTROL-PLANE WRITE instant, not the agent's collection instant:
+    -- the metadata payload carries no collection timestamp, so that fact is not
+    -- knowable here. Hence "updated_at" and not "collected_at".
+    --
+    -- Written ONLY by UpdateSiteMetadata (db/query/sites.sql), the only
+    -- statement in the tree that writes components, and written on every push
+    -- even when the document is byte-identical (freshness is when the claim was
+    -- last confirmed, not when it last changed). A disconnecting site is left
+    -- alone: the stamp goes stale and the staleness IS the signal. Nulling it
+    -- would erase a fact that happened while leaving components itself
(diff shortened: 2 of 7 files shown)


mosamlife:
## Gates

Each run unpiped, with its own exit status.

| Gate | Exit |
|---|---|
| `go build ./...` (from `apps/api`) | 0 |
| `go vet ./...` (from `apps/api`) | 0 |
| `go test ./internal/...` (from `apps/api`) | 0 |
| `sqlc generate` (`/Users/…/go/bin/sqlc`, resolved with `command -v`) | 0 |
| `git diff --quiet -- internal/db/sqlc/` | 0 |
| `make gen` | 0 |
| `git status --porcelain` after `make gen` | empty |
| `make test-integration` | 2 — one failure, infrastructure, see below |

`sqlc` resolved with `command -v sqlc`; installed version `v1.31.1`, matching the tree's stamp (`grep -h 'sqlc v' apps/api/internal/db/sqlc/*.go | sort -u`). The generated tree was regenerated, never hand-edited, and the empty `git diff --quiet` above is the proof that what is committed equals what the generator produces.

### The integration failure

`make test-integration` (which CI does not run — `ci.yml` excludes `apps/api/tests`) exited 2 with exactly one failure:

```
--- FAIL: TestGH402_KindCheck_DoesNotBlockTheSiteDeleteEnqueue (12.41s)
    gh402_reclaim_kind_check_integration_test.go:158: SETUP FAILURE (infrastructure, not the test's own assertion)
        at stage="postgres: connection string": port "5432/tcp" not found
```

A testcontainers port-binding flake under parallel load, in a test about m115's reclaim kind check — a different table, and it never reached an assertion. Not taken on trust: re-run in isolation, it passes.

```
$ go test -count=1 -timeout 20m -run 'TestGH402_KindCheck_DoesNotBlockTheSiteDeleteEnqueue' ./tests/
ok  	github.com/mosamlife/wpmgr/apps/api/tests	3.219s      # exit 0
```

Counts from the run log, each with its command:

```
$ grep -cE '^--- FAIL:' <log>                         # 1
$ grep -cE 'SETUP FAILURE \(infrastructure' <log>     # 1
```

One failure, one infrastructure setup failure — the same one. The other three packages were green: `tests/contract`, `tests/gh458`, `tests/testinfra`.

PR #561, opened 2026-08-27
fix(web): honest inventory-age-unknown state for null as_of (GH #553)

mosamlife:
## Summary

- `GET /sites/{id}/updates/available` now reports `as_of: null` (from `sites.components_updated_at`) for every site that existed before the GH #553 fix, until its next inventory sync. `updates/refresh` schema in `packages/openapi/openapi.yaml` documents this explicitly: "Explicit null when the inventory has never been collected; never falls back to the site's general updated_at."
- The available-updates card rendered that `null` through the shared `FreshnessBadge`, which shows the literal text "Never" for a null `collectedAt`. That recreates the exact bug GH #553 fixed, one layer up: the returned `items`/`core_update` are real, already-collected data, and "Never" wrongly asserts the inventory was never collected.
- Both places `as_of` is rendered (the card header's freshness line, and the "All managed components are up to date" empty state) now show a dedicated "Inventory age unknown" state instead, with plain-language copy on what resolves it: the site's next inventory sync, usually within 30 minutes. A real `as_of` timestamp still renders through the unmodified `FreshnessBadge` ("Updated Xm ago" / "Stale").
- `FreshnessBadge` itself is untouched: its two other call sites (`health-tab.tsx`, `diagnostic-card.tsx`) key off `collected_at` on individual diagnostic cards, where `null` genuinely does mean "never collected," so its "Never" copy stays correct there.

## Before vs after

- Before: `as_of: null` -> `FreshnessBadge` -> renders "Never" (confirmed by adding a test against the pre-fix code and watching it fail; see test plan).
- After: `as_of: null` -> "Inventory age unknown. Clears after this site's next sync, usually within 30 minutes." Copy contains no raw field names (`as_of`, `components_updated_at`) and no error codes.

## Test plan

- [x] Added a test asserting the pre-fix behavior is gone (the literal "Never" no longer appears) and the new copy is present and says what resolves it, using `as_of: null`.
- [x] Added a test asserting the unknown-age copy is absent when a real `as_of` timestamp is supplied.
- [x] Ran the first test against the pre-fix code and watched it fail (red), confirming it was actually exercising the bug, then watched it pass after the fix (green).
- [x] `pnpm -C apps/web typecheck` — exit 0
- [x] `pnpm -C apps/web lint` — exit 0 (0 errors, 9 pre-existing warnings unrelated to this change)
- [x] `pnpm -C apps/web test` — exit 0 (145 files, 1730 tests)
- [x] `impeccable detect` on the touched file — clean (`[]`)

Not merging; opening for review per standing repo policy that the only route to `main` is a PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved update freshness messaging when inventory age is unavailable.
  * Replaced unclear “Never” or raw field displays with guidance that the status clears after the next site sync, typically within 30 minutes.
  * Preserved freshness details when a valid collection timestamp is available.

### DIFF
--- a/apps/web/src/features/updates/available-updates-card.test.tsx
+++ b/apps/web/src/features/updates/available-updates-card.test.tsx
@@ -201,3 +201,35 @@ describe("AvailableUpdatesCard agent honesty (GH #314)", () => {
     expect(within(list).queryByText("WPMgr agent")).not.toBeInTheDocument();
   });
 });
+
+// GH #553: `as_of` used to be stamped from the 60s site heartbeat, so it was
+// practically never null. It now reads `components_updated_at`, which is only
+// set when the inventory is genuinely (re)written, and it shipped with no
+// backfill, so every site that existed before the fix reports `as_of: null`
+// until its next sync. `null` is a real, common, temporary state here, not an
+// absence to paper over with a generic "Never".
+describe("AvailableUpdatesCard as_of honesty (GH #553)", () => {
+  it("tells the operator the collection time is unknown, and how it resolves, instead of claiming it was never collected", () => {
+    setup(payload({ as_of: null }), null);
+    renderWithProviders(<AvailableUpdatesCard siteId="site-1" />);
+
+    // Distinct from FreshnessBadge's generic "Never", which would wrongly
+    // assert this site's inventory was never collected -- it was, we just
+    // don't have a genuine collection time for it yet.
+    const unknownCopy = screen.getAllByText(/inventory age unknown/i);
+    expect(unknownCopy.length).toBeGreaterThan(0);
+    expect(screen.queryByText(/^Never$/)).not.toBeInTheDocument();
+
+    // States what resolves it, in operator terms, with no raw field names.
+    expect(screen.getAllByText(/next sync/i).length).toBeGreaterThan(0);
+    expect(screen.queryByText(/as_of/i)).not.toBeInTheDocument();
+    expect(screen.queryByText(/components_updated_at/i)).not.toBeInTheDocument();
+  });
+
+  it("does not show the unknown-age copy when a real collection time is present", () => {
+    setup(payload({ as_of: "2026-08-04T10:00:00Z" }), null);
+    renderWithProviders(<AvailableUpdatesCard siteId="site-1" />);
+
+    expect(screen.queryAllByText(/inventory age unknown/i).length).toBe(0);
+  });
+});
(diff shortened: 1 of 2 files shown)


mosamlife:
Resolved in 2fec2df0.

**Thread 1 (unavailable vs. unknown).** Confirmed real: the header freshness line rendered unconditionally, outside the `isPending`/`isError` branch, so `data?.as_of ?? null` fell through to `null` for BOTH a still-loading query and a failed fetch, and both then showed "Inventory age unknown. Clears after this site's next sync" -- wrong in each case (a loading query has established nothing yet; a failed fetch never loaded the inventory at all, so a sync fixes nothing).

`UpdatesAsOf` now takes `isPending`/`isError` and has three states:
- loading: renders nothing (no claim to make yet)
- failed (`isError`): "Inventory age unavailable. The update check did not load." -- distinct copy, no raw field names, no error code. The existing `PageError` body ("Could not load available updates") still owns the *why*; this line only says an age can't be shown.
- unknown (`asOf === null`, request succeeded): unchanged "Inventory age unknown. Clears after this site's next sync, usually within 30 minutes."

Proof: added a test for the loading case and one for the error case, both asserting the wrong copy is absent and (for error) the new copy is present. Both failed against the pre-fix code (confirmed via `pnpm -C apps/web test`, 4 failing), then passed after.

**Thread 2 (duplicate).** Confirmed real, not hypothetical -- verified by rendering the null-`as_of`/zero-updates case and inspecting the DOM: "Inventory age unknown..." appeared twice, once in the header and once inside the "all managed components are up to date" empty-state block, because the header renders unconditionally and the empty state rendered its own copy of the same call whenever `total === 0`. The header is the only location that renders in every query state (loading/error/loaded), so it now owns the claim alone; the empty-state's duplicate render was deleted outright rather than suppressed conditionally. Added a regression test asserting exactly one occurrence (`toHaveLength(1)`), which failed (2 found) against the pre-fix code and passes now.

**Failed-fetch path, checked as asked.** Before this commit it landed on the same freshness-badge problem as thread 1 (see above) -- it showed "Inventory age unknown...next sync" during `isError`, same as loading. Fixed by the same change.

Gates on `2fec2df0` (branch rebased on origin/main eba7288a):
- `pnpm -C apps/web typecheck` -> exit 0
- `pnpm -C apps/web lint` -> exit 0 (0 errors, 9 pre-existing warnings, none in touched files)
- `pnpm -C apps/web test` -> exit 0 (145 files, 1733 tests)
- `impeccable detect --json src/features/updates/available-updates-card.tsx` -> `[]`, exit 0

PR #582, opened 2026-08-29
test(auth): pin the address the auth rate limiters key on

mosamlife:
## What this is

Three regression pins for **which address the two decision-making auth rate limiters key on**:

| Limiter | Key | Cap |
|---|---|---|
| `internal/auth/reset.go` | `"pwreset-consume:"+ip` | `resetPerMinute` |
| `internal/auth/twofa.go` | `"2fa-ip:"+ip` | `twoFAIPLockoutPerMinute` |

Both derive that address from `clientAddr()` → `gin.Context.ClientIP()`. What `ClientIP()` resolves to is a function of how the engine is configured, and **nothing in the suite pinned it**. These pins state the requirement: one peer gets one bucket, regardless of what that peer sends.

## Expected red

**These pins fail on this branch and are expected to.** They are landing ahead of the remedy deliberately — their absence is why the derivation was never constrained. Do not merge to green them by weakening them.

Each pin runs an honest control arm in the same process, and the control asserts in **both** directions:

- past the cap the limiter must still refuse (a limiter that stopped refusing fails `CONTROL BROKEN`);
- below the cap it must still serve (a limiter that refuses legitimate traffic fails `CONTROL OVER-FIRES`).

That second assertion matters, because the obvious remedy has a failure mode where legitimate traffic gets bucketed together. The pins will catch that rather than reward it.

Satisfiability was checked before landing: under a candidate remedy the varied arm collapses to exactly the control (`10/30` and `30/90`) — so the pins are passable and are not asserting something unreachable.

## Why the remedy is not in this PR

Keying these correctly requires one deployment fact: **how many proxy hops sit in front of the process.** That fact is not in this repository — `infra/cloudbuild.api.yaml` deliberately carries no environment-specific values, and there is no load-balancer, url-map, backend-service or forwarding-rule config anywhere in `infra/` or `.github/workflows/`.

Both candidate remedies need it, and both have a bad failure mode if it is guessed wrong:

- declaring the trusted proxies — correct everywhere at once, but wrong values break every genuine client address;
- keying on the peer address directly — unspoofable regardless of topology, but if a proxy sits in front then every request shares that proxy's address and the limiter buckets all traffic into one key, which refuses legitimate users.

Guessing between these is how you ship something that looks like a fix and is not, or an outage. So the pins land first and the remedy follows once the hop count is established. **The evidence gathered so far, and the single command that settles it, is in the review thread.**

## Checks

- `go build ./...` — pass
- `go vet ./...` — pass
- `go test ./internal/auth/` — **3 failures, all three of these pins, nothing else in the package**

## Not in scope

- `infra/nginx/nginx.conf` has the same shape for self-hosted installs — `devops-engineer`'s path, being routed separately.
- `internal/rum/handler.go` derives its address by an approach that cannot be reconciled with the remedy here; flagged for a separate fix, see the review thread.
- `autologin/ratelimit.go`'s `MemoryLimiter` is per-process, so every limit is already multiplied by the instance count. Independent of this.

Draft. `security-reviewer` pass to follow.


mosamlife:
Closing for now — reopening with the fix once the deployment topology is confirmed. Work is preserved locally.

PR #594, opened 2026-08-30
fix(mcp): apply org-scope enforcement to the MCP consent routes

mosamlife:
An MCP grant is an org-level object: `site_scope_mode: 'all'` resolves to every site in the organisation at read time, so authorizing one is an org-member action rather than a per-site one. These routes carry no `:siteId` for `RequireSiteAccess` to bind, which makes `RequireOrgScope` the applicable gate — the same reasoning as the update-run orchestrator and the fleet rollups.

## Change

Three layers, each independently sufficient:

- `authz.RequireOrgScope()` on the `/authorize` and `/consent` group, applied inside `Handler.Register` so every mount carries it — including the integration harness. A gate that exists only in `server.New` is a gate no test exercises.
- `Service.Approve` restates it, so a caller that reaches the service without the route middleware is refused too.
- `Repo.CreateGrantWithCode` mints through `db.RunTenantTx` rather than choosing a tx helper itself. `db.go` documents `RunTenantTx` as mandatory for repos precisely so a call-site cannot fall outside the site-scope RLS; only it can route a site-constrained principal into the transaction where the RESTRICTIVE policy on `mcp_grants` is live.

The middleware alone would leave that third layer unpinned: it is one edit from being absent, and below the SQL layer nothing would notice its absence.

`ApprovalRequest` now carries the whole `domain.Principal` instead of a `(TenantID, UserID)` pair — the pair could be filled in correctly while `Scope` and `AllowedSiteIDs` were silently dropped.

## Tests

`apps/api/tests/mcp_consent_org_scope_integration_test.go` drives the **mounted routes** as `wpmgr_app`, not against a fake, and asserts both halves:

- a site-scoped collaborator is refused **and leaves no row** (a 403 that committed first is not a refusal);
- an ordinary org member still completes consent and receives a code.

Service-level counterparts in `internal/mcp/approve_test.go`, including the fail-closed backstop cases. `mountLikeProduction` now takes a full principal, so scope is expressible from the harness at all.

Also corrects the comment on `Repo.CreateGrantWithCode`, which asserted a guarantee this call path did not have.

## Checks

`go build ./...`, `go vet ./...`, `go test ./internal/... ./cmd/...` all clean, unpiped. `make test-integration` status reported in the review thread.

Draft: `security-reviewer` pass to follow before merge.

## Summary by CodeRabbit

* **New Features**
  * Added organization-level access enforcement for MCP OAuth authorization and consent.
  * Organization members can complete consent and receive authorization codes.
* **Bug Fixes**
  * Site-scoped access is rejected with HTTP 403 and an OAuth-standard `access_denied` response.
  * Requests without an authenticated principal return HTTP 401.
  * Prevented unauthorized grant creation and ensured grants honor access boundaries.
* **Tests**
  * Added coverage for permitted, denied, and database-enforced grant flows.

### DIFF
--- a/apps/api/internal/mcp/approve_test.go
+++ b/apps/api/internal/mcp/approve_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/google/uuid"
 
 	"github.com/mosamlife/wpmgr/apps/api/internal/db/sqlc"
+	"github.com/mosamlife/wpmgr/apps/api/internal/domain"
 )
 
 // ---------------------------------------------------------------------------
@@ -49,10 +50,17 @@ func validConsent() ConsentContext {
 	}
 }
 
+// validApproval is an ORG-scoped approver: Scope "org" and no allowlist, which
+// is what every pre-existing case in this file assumes. The site-scoped
+// counterpart is exercised by TestApprove_RefusesASiteScopedApprover below and
+// end to end in tests/mcp_consent_org_scope_integration_test.go.
 func validApproval() ApprovalRequest {
 	return ApprovalRequest{
-		TenantID:  uuid.New(),
-		UserID:    uuid.New(),
+		Principal: domain.Principal{
+			TenantID: uuid.New(),
+			UserID:   uuid.New(),
+			Scope:    domain.ScopeOrg,
+		},
 		Consent:   validConsent(),
 		GrantName: "Claude Desktop on my laptop",
 		SiteScope: SiteScopeRequest{Mode: SiteScopeModeAll},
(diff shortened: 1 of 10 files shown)


mosamlife:
## Re: "OAuth denial envelope is bypassed"

**The finding's stated consequence is wrong, but it found a real defect — and the defect is worse than described.** Fixed in `31efdc3b`.

### Does any OAuth client read these responses? No.

Traced both routes. Neither is consumed by an OAuth client:

- `authorization_endpoint` is `/api/v1/oauth/mcp/authorize` (`internal/mcp/discovery.go:38`), and that file's own comment says a client "opens a browser at authorization_endpoint and posts to token_endpoint" — the browser navigates, the client does not parse.
- Both routes are called by **our own dashboard** via raw same-origin `fetch`: `apps/web/src/features/mcp-consent/use-consent.ts:27-28` defines `CONSENT_AUTHORIZE_PATH` / `CONSENT_APPROVE_PATH`, fetched at `:130` (GET) and `:171` (POST). That file explains why it bypasses the generated client: these endpoints are not in the OpenAPI contract.
- An OAuth client learns of a refusal from a **redirect**, not a body — `buildDenialTarget` sends `error=access_denied` plus the original `state` to the registered redirect URI (`features/mcp-consent/denial.test.ts`).

So no OAuth client is "prevented from parsing" anything.

### But the envelope inconsistency was real, and it is mine

`use-consent.ts:62` states the contract outright: *"every code below is one `oauthError()` in `apps/api/internal/mcp/dto.go` emits."* `readOAuthError` (`:74-91`) reads `error` and `error_description` and **defaults `code` to `"server_error"`** when `error` is absent.

`RequireOrgScope` aborts with `{code, message}`. So a site-scoped collaborator following an AI client's link got the screen's **`server_error`** branch — the system reporting itself broken — for a deliberate, correct refusal. Not merely generic: actively misleading.

Executed, with the middleware reverted to `authz.RequireOrgScope()`:

```
POST /consent refused with error="", want "access_denied"
POST /consent answered in the generic {code, message} envelope (code="org_scope_required")
GET  /authorize refused with error="" code="org_scope_required"
```

### The fix — and it does not weaken the refusal

A package-local `requireOrgScope()` using the **same predicate** (`domain.Principal.IsSiteConstrained`, the single definition every other site gate calls — no second opinion about who is constrained), emitting through `oauthError()` so status and envelope are still decided in exactly one place.

**Status is unchanged at 403** via `ErrCodeAccessDenied`. I did not trade a correct status for a prettier body — the mutation above shows the status stayed 403 in both shapes; only the envelope differed. 401 would invite a credential retry that cannot succeed.

After:

```
POST /consent refused the collaborator with 403 (error="access_denied" desc="site-scoped access cannot authorize an organisation-level connection")
GET  /authorize refused the collaborator with 403 (error="access_denied")
```

Tests now assert the envelope, not just the status, on both routes; reverting the middleware reddens them naming the generic code.

### `apps/web` — untouched, and nothing needed there

No change is required on the screen: it already parses the OAuth envelope correctly and already renders per-code branches. The bug was the server sending a shape the screen documents it will not receive. Nothing here collides with #592.

### Carried into the code

Added at the mount, since it explains why both routes needed the gate rather than just the writing one:

> `/consent` does not require a prior `/authorize` — Approve re-resolves the client and re-matches the redirect from the POST body, and `/register` is unauthenticated, so the write is reachable with a self-registered client and a single POST: no consent screen, no operator interaction.

### Checks

`go build ./...`, `go vet ./...`, `go test ./internal/... ./cmd/...` — exit 0, unpiped, after the rebase onto `main`. All 12 `TestMCP*` integration tests green as `wpmgr_app` (`rolsuper=false rolbypassrls=false`). Full `./tests/...` re-running on the rebased tree; I will report the status rather than assume it.

Not merging — yours.

PR #609, opened 2026-08-30
feat(mcp): mint a connection token from the dashboard, the headless path

mosamlife:
Adds `POST /api/v1/mcp/connections` — the non-OAuth path the wizard needs.

No MCP client documents a device-code flow and Claude Code cannot run browser sign-in non-interactively, so headless, SSH and CI installs have no route through `/authorize` + `/consent`. This is that route: one transaction creates the grant with the operator's name, site scope and capabilities, mints its first connection token, and returns the plaintext exactly once.

## What it reuses rather than rebuilds

It goes through the same grant-creation shape as consent, so m127's `capabilities` / `expires_at` columns and the `mcp.grant.created` audit event come with it. The credential is built by the existing `hashCredential` (lower-case hex SHA-256) and stored by the existing `CreateMCPConnectionToken`. The only thing that differs from the OAuth path is how the grant is authorised into being.

Audit metadata carries an `issuance` discriminator so a headless mint is distinguishable from a browser-consented one, plus the public `token_prefix` — never the plaintext.

## The org-scope guard

Applied at all three layers the consent path carries it at:

1. `authz.RequirePermission(authz.PermAPIKeyManage)` on the route — an `orgLevelPerms` member, so `RequirePermission` refuses any site-constrained principal outright.
2. `requireOrgScopedPrincipal` in the service.
3. `mcp_grants_site_scope_insert` in the database, reached because `CreateGrantWithToken` takes the principal and runs `RunTenantTx` rather than picking `InTenantTx` itself.

## Tag ids, not names — and why the refusal is loud

The wire takes tag and site **UUIDs**, matching `approvalRequestDTO.ScopeTagIDs`, so there is one vocabulary for a tag across both mint paths and the wizard reuses the consent screen's existing name→id map. Ids also survive a rename; a grant scoped by name would silently change which sites it covers the day somebody renames a tag.

`scope_tag_ids` and `scope_site_ids` are `uuid[]` with no foreign key, so an id naming nothing stores cleanly and then resolves to zero sites forever. Those ids are therefore verified against the tenant's own registry and refused **by id**.

That is deliberately **not** the same question as "does the scope resolve to any sites":

| input | outcome |
|---|---|
| real tag, zero sites today | accepted, stored as given |
| id naming no tag at all | refused, loudly, by id |

An empty scope is legitimate and is stored as empty, never widened to `all`. The structural refusals are unchanged — an empty allowlist is still not a way to ask for every site.

## Rate limit

Reuses `register_limit.go`'s two-layer bucket. The key is the session-resolved user id rather than `RemoteIP`: this endpoint is authenticated so the id cannot be set by a header, and the harm being bounded (one stolen session becoming many long-lived credentials) is a per-user quantity.

## Capabilities

An omitted list means the organisation default, never an empty set — an empty stored set is a connection `Authenticate` refuses on every request, so minting one would produce a credential that can never work. A named capability wider than the ceiling is refused rather than dropped.

## Status

Draft. Unit proofs, mutation sweep and `make test-integration` reported in the thread.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/api/internal/mcp/dto.go
+++ b/apps/api/internal/mcp/dto.go
@@ -187,6 +187,74 @@ type connectionListDTO struct {
 	Connections []connectionDTO `json:"connections"`
 }
 
+// mintConnectionRequestDTO is the wizard's headless-connection request.
+//
+// THE TAG FIELD IS `scope_tag_ids` AND IT CARRIES UUIDs, deliberately spelled
+// the same as approvalRequestDTO.ScopeTagIDs (above) so there is ONE wire
+// vocabulary for a tag across both mint paths. The consent screen already holds
+// the name->id map and refuses its own submit when the tag registry went null
+// mid-flight; the wizard reuses that map rather than introducing a name form
+// this endpoint would have to disambiguate on case and duplicates.
+//
+// Ids also survive a rename and names do not: a grant scoped by name would
+// silently change which sites it covers the day somebody renames a tag, and
+// nothing would say so. The server-side half of the bargain is that an id
+// naming no tag is REFUSED by id rather than dropped -- see
+// ErrCodeUnknownScopeTag.
+type mintConnectionRequestDTO struct {
+	Name          string   `json:"name"`
+	SiteScopeMode string   `json:"site_scope_mode"`
+	ScopeTagIDs   []string `json:"scope_tag_ids"`
+	ScopeSiteIDs  []string `json:"scope_site_ids"`
+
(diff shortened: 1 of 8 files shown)


mosamlife:
Merged into `main` as part of #615 rather than directly — `git merge-base --is-ancestor origin/worktree-agent-a14d06d8f125c882a origin/main` passes, so every commit here is on `main`. Closing to keep the list honest; nothing was dropped.

PR #614, opened 2026-08-30
merge: reconcile m128 setup_client onto the m127 expiry/capabilities stack

mosamlife:
Reconciles m128 (#610) onto the MCP backend stack. This is a merge resolution, not new schema design — no migration file was written or edited here.

## Apply order

m127 is `20260829000000`, m128 is `20260830000000`, so m128 applies second and no reordering was needed. The two files commute as m128's author predicted: neither reads the other's columns and neither backfill touches the other's rows. **No converge path is owed.** Each migration's own backfill is already `WHERE ... IS NULL`-guarded and re-runnable, and a database that has applied either one in isolation reaches the same end state when the other arrives.

## What conflicted

`db/schema.sql` auto-merged and was verified by hand to carry both column sets. The one real conflict was `CreateMCPGrant` in `db/query/mcp_connections.sql`, where both sides extend the same INSERT column list. Both sides' prose was kept and the column list carries all four new columns — twelve columns against twelve placeholders, `RETURNING *` unchanged.

The three `internal/db/sqlc/` files were resolved by taking one side to clear the markers and then **regenerating**; nothing in that tree was hand-edited.

```
/Users/mosamgor/go/bin/sqlc  ->  v1.31.1
grep -h 'sqlc v' apps/api/internal/db/sqlc/*.go | sort -u  ->  //   sqlc v1.31.1
```

Regeneration is pure: a second identical `sqlc generate` did not move the tree (`git diff --quiet -- internal/db/sqlc/` exit 0).

## Nullability, which is the point of the merge

The split survives and is asymmetric on purpose. `capabilities` and `expires_at` stay NOT NULL, so an omitting caller gets 23502 rather than an unrestricted or never-expiring connection. `setup_client` and `idle_expire_after_days` stay nullable, because absence is a real value on both — `SetupClient *string` and `IdleExpireAfterDays *int32` against `Capabilities []string` and `ExpiresAt time.Time` in the generated params struct.

## Finding: nothing in the fast lane guards m128's column

A mutation sweep dropped each side's columns from the resolved INSERT in turn.

Dropping m127's three reddens immediately and by name, at six call sites in `internal/mcp/mint.go` and `internal/mcp/service.go` (`unknown field Capabilities`, `ExpiresAt`, `IdleExpireAfterDays`).

**Dropping `setup_client` compiles clean and `go test ./internal/mcp/...` stays green.** The cause is that m128's Go wiring does not exist yet on this stack:

```
grep -rn "SetupClient" apps/api --include="*.go" | grep -v "internal/db/sqlc/"
  -> no matches
```

So the column, the query and the generated field all exist, and no Go code populates any of them. `apps/api/tests/mcp_m127_m128_merge_roundtrip_integration_test.go` is the guard that was missing: one test supplies both column sets and asserts every value round-trips, the other omits `setup_client` and asserts NULL. Both run as `wpmgr_app` through `RunTenantTx` — the only helper that sets `app.site_scope`, so the RESTRICTIVE `mcp_grants_site_scope_*` policies are actually bound rather than inert — and both assert `current_user` / `rolsuper` / `rolbypassrls` inside the transaction that runs the INSERT.

## Proofs

Each status is from an unpiped run.

- `go build ./...`, `go vet ./...`, `go vet ./tests/` — exit 0.
- `go test -count=1 ./internal/...` — exit 0, and `grep -c 'FAIL'` over the captured output returns 0.
- `scripts/check-rls-cross-tenant_test.sh` — exit 0, `54 passed, 0 failed`.
- `scripts/check-rls-cross-tenant.sh` — exit 0. It built its **own throwaway container** (`wpmgr-rls-guard-48585` on `127.0.0.1:49566`, 137 migration files applied); `WPMGR_RLS_DATABASE_URL` was confirmed unset beforehand and the local dev database was never read. Extracted 260 policies, 105 cross-tenant, checked against 105 ledger rows.

### Not yet run

`make test-integration` **has not completed.** Launched from this worktree's root, it compiled the `tests` binary and then blocked on the machine-wide lock, which is held by another worktree (`agent-ace6ad5e8ff381e77`) running the same target. It should be re-run before merge; the two new tests in this branch have never been executed, only compiled.

An earlier `go test ./internal/...` run printed `FAIL` with no failing package in the captured tail. A `-count=1` run and two cached runs were all clean. That earlier failure is **unexplained and unreproduced**, not diagnosed as a fluke.

### DIFF
--- a/apps/api/db/query/mcp_connections.sql
+++ b/apps/api/db/query/mcp_connections.sql
@@ -268,11 +268,22 @@ RETURNING *;
 -- either gets 23502 rather than a live organisation-wide grant (Decision 1).
 -- The caller passes status explicitly -- 'active' -- rather than relying on the
 -- schema to assume it.
+-- setup_client (m128) IS THE OPERATOR'S CHOICE AT S29 STEP 2 and is passed
+-- explicitly, like status, rather than being left to the schema. It is NULLABLE
+-- with NO DEFAULT precisely so a caller that never asked -- any path that is
+-- not the step-2 wizard -- can pass NULL and mean "no operator choice was
+-- recorded". PASS NULL RATHER THAN 'generic' ON SUCH A PATH: 'generic' asserts
+-- the operator saw nine cards and chose "Other MCP client", which is a
+-- different fact and the one S29 step 9 distinguishes.
+--
+-- Do NOT derive it from client_name. That column is self-reported at
+-- `initialize`, is NULL until the client first connects, and inferring a
+-- choice from it manufactures a fact the operator never stated.
 INSERT INTO mcp_grants (
     tenant_id, name, status, site_scope_mode, scope_tag_ids, scope_site_ids,
-    client_id, created_by_user_id
+    client_id, created_by_user_id, setup_client
 ) VALUES (
-    $1, $2, $3, $4, $5, $6, $7, $8
+    $1, $2, $3, $4, $5, $6, $7, $8, $9
 )
 RETURNING *;
 
--- a/apps/api/db/schema.sql
+++ b/apps/api/db/schema.sql
@@ -5651,6 +5651,36 @@ CREATE TABLE mcp_grants (
     client_version              text        NULL,
     protocol_version            text        NULL,
     client_identity_recorded_at timestamptz NULL,
+    -- m128. THE OPERATOR'S CHOICE at S29 step 2, as a client-table.ts slug.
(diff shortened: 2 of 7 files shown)


mosamlife:
Merged into `main` as part of #635 rather than directly — `git merge-base --is-ancestor` confirms every commit on this branch is on `main`. #635 carried the m128 reconciliation together with the Go wiring for `setup_client`, which is what the column was waiting for. Closing to keep the list honest; nothing was dropped.

PR #615, opened 2026-08-30
fix(mcp): attribute grant approve and revoke to the credential that acted

mosamlife:
Closes the two remaining instances of the nil-user audit-attribution defect
that #613 fixed for `MintConnection`.

## The defect

`Service.Approve` and `Service.RevokeConnection` both hardcoded
`ActorType: audit.ActorUser` beside `ActorID: Principal.UserID.String()`.
`apikey.PrincipalFor` sets `APIKeyID` and never `UserID`, so an API-key caller
wrote `uuid.Nil` into `actor_id` under an `actor_type` asserting a human acted.
The row exists, hashes into the chain, and explains nothing: `actor_id` names no
user, and because `ListAuditEntries`' name join is gated on `actor_type`
(`'user'` -> `users.name`, `'api_key'` -> `api_keys.name`) the row resolves to no
name either.

Both routes are reachable by an API key today:

- **revoke** — `server.go`'s `RegisterConnections(v1)` mounts it from the same
  nil check as the headless mint. A key that can mint can revoke, with nothing
  in between. This is the row an auditor reaches for after an incident.
- **consent** — `server.go`'s `Register(v1)` mounts `/consent`, and `v1` derives
  from `sessionAuthGroup`, which carries `Auth.Authenticate()` and its Bearer
  branch. Verified by reading the mount, not assumed: browser-only consent is a
  convention about who usually calls it, not a constraint on who can.

## The resolver moved

`mintAuditActor` moved out of `internal/mcp/mint.go` into `internal/audit` as
`audit.ActorFor`. Three callers now share it, the concept is not mint-specific,
and it belongs beside the actor constants and the actor-name join whose two arms
its returned pair must agree with. `audit` already imports `domain`, so the
direction is cycle-free. There is still exactly one resolver.

`CreatedByUserID` is unchanged and still NULL for a key.

## Prose corrected

`internal/audit/audit.go` said `mcp.grant.revoked` is written by `ActorUser`
with "their user id", and that `mcp.grant.created` is `ActorUser` on the consent
path. This change makes both false; corrected. Also dropped `grant_name` from
the documented metadata for `mcp.grant.revoked`, which `RevokeConnection` does
not write.

## Proofs

`apps/api/tests/mcp_audit_actor_attribution_integration_test.go` drives each
fixed site with an API-key principal through the mounted handler and reads the
audit row back as `wpmgr_app` through the same `InTenantTx` dispatch the request
uses. `mcpAssertAndReportRole` prints `current_user`/`rolsuper`/`rolbypassrls`
from `pg_roles` inside that transaction. Every assertion is by value — the
`actor_type` and `actor_id` expected, named — never merely that a row was
recorded. Each site is paired with a session user driving the same route, so an
inverted fix cannot pass either half alone.

Mutation sweep: reverting each of the three call sites to the old
`ActorUser`/`UserID` pair in turn reddens the proof belonging to that site, and
only that one.

## Summary by CodeRabbit

* **New Features**
  * Added API support for creating MCP connection tokens with site/tag scopes and selected capabilities.
  * MCP connections now support capability restrictions, absolute expiration, and optional idle expiration.
  * Connection activity is tracked to keep active connections from idle-expiring.
* **Security**
  * Tokens are shown only once and stored securely.
  * Added validation, rate limits, tenant-scope checks, and permission controls.
* **Audit**
  * Added audit records for connection creation, tool calls, and revocation.

### DIFF
--- a/apps/api/cmd/dump-routes/routes.go
+++ b/apps/api/cmd/dump-routes/routes.go
@@ -286,7 +286,7 @@ func buildEngine() (engine *gin.Engine, omittedDepsFields []string, err error) {
 	// cmd/wpmgr/main.go wires it (mcp.NewService(mcp.NewRepo(pool)) etc.), so
 	// POST /mcp, the four OAuth paths and the three well-known discovery
 	// documents all mount.
-	mcpSvc := mcp.NewService(mcp.NewRepo(pool)).WithClock(clock.Now)
+	mcpSvc := mcp.NewService(mcp.NewRepo(pool)).WithClock(clock.Now).WithAudit(auditRec)
 	mcpTransportH := mcp.NewTransportHandler(mcpSvc, logger, "dump-routes")
 	mcpOAuthH := mcp.NewHandler(mcpSvc)
 	mcpDiscoveryH := mcp.NewDiscoveryHandler("https://cp.example.test")
--- a/apps/api/cmd/wpmgr/main.go
+++ b/apps/api/cmd/wpmgr/main.go
@@ -540,7 +540,7 @@ func run(ctx context.Context, cfg config.Config, logger *slog.Logger) error {
 	// would make POST /mcp answer 404, which is indistinguishable from a
 	// routing failure. Self-host reaches it too; the grant is what authorizes,
 	// and an installation with no grants simply has no valid bearer token.
-	mcpSvc := mcp.NewService(mcp.NewRepo(pool)).WithClock(clock.Now)
+	mcpSvc := mcp.NewService(mcp.NewRepo(pool)).WithClock(clock.Now).WithAudit(auditRec)
 	mcpTransportH := mcp.NewTransportHandler(mcpSvc, logger, version)
 	// The OAuth half, wired unconditionally for the same reason. Without it the
 	// transport above is mounted and correct and refuses every request forever,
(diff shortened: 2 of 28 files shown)


mosamlife:
Both bot findings are addressed in this branch, and the stale-comment finding turned out to be wider than either bot saw.

**`domain.Unauthorized` -> `domain.Forbidden`** at the empty-capability refusal in `Authenticate`. Verified first that every other producer of `ErrCodeCapabilityUnmapped` already returns `Forbidden` (three sites in `policy.go`), and that this was the only `Unauthorized`. The test that let it through asserted only that an error occurred; it now asserts `Kind`, HTTP status and code by value, and reddens when the change is reverted.

**The audit-rollback probe** is pinned with `errors.Is(err, errMintAuditProbe)`. Reverting the pin alone reddened nothing, so the failure it exists to catch was planted instead — an invalid `site_scope_mode` aborting the insert before `onCreated` runs. Red with the pin, green without it. The vacuous pass was executed rather than argued.

**The stale `TouchMCPGrantInTenantTx` comment** was in two source sites, not one, feeding three generated surfaces rather than two: a doc comment lands in both generated files while an in-body comment lands only in the const string, so `querier.go` was invisible to a diff-based read. The conclusion is kept and the reason replaced — the idle column stays NULL because nothing asks the operator for a window yet, not because the stamp is missing. `schema.sql` corrected too, as a current-state document.

The m127 migration text is deliberately left carrying the old claim. It is not on `main` yet so it is still editable today, and that is a choice rather than an inability: the text was true when written, `schema.sql` carries the corrected version, and a dated record a reader can place beats history rewritten to look prescient. Please do not "fix" it later.

Verified as a combination, not per-branch: this backend tree is byte-identical under `apps/api` to the tree that ran `make test-integration` green (`INTEGRATION_EXIT=0`, 4 packages ok, 0 failures, 864s), together with the frontend PR.

PR #644, opened 2026-08-31
feat(db): m132 restrictive site-scope RLS on the assistant read surface

mosamlife:
ADR-061 A11 Wave 1.2. Adds the RESTRICTIVE `app.site_scope` policy to 22 site-keyed tables the assistant reads, using the m19 `sites_site_scope` predicate byte for byte.

## Scope

Migration + `db/schema.sql` only. No Go. `db/query/**` unchanged (no column or query change), so the sqlc tree is byte-identical after regeneration.

## What it buys, and what it does not

Two of the 22 are read **today**: `site_perf_config` and `site_object_cache_config` are LEFT JOINed by `db/query/sites.sql`, the query behind `list_sites` — the only tool in the registry. The rest become reachable as the tool registry widens inside the m131 capability vocabulary.

The policies are **inert** on every path that does not set `app.site_scope`, which today is every path except `InScopedTenantTx` (`internal/db/db.go:582`). That is stated in the migration rather than left to assumption. Wave 1.3 changes `ResolveScopeSites` (`internal/mcp/repo.go:494`) to take a principal and is what makes them load-bearing. **A scoped helper cannot engage policies that do not exist, so this lands first.**

## A11 item 3 — empty allowlist fails closed

A property of the predicate, not of the application:

```
nullif(<empty string>, <empty string>)  -> NULL
string_to_array(NULL, ',')              -> NULL
site_id = ANY (NULL::uuid[])            -> NULL
FALSE OR NULL                           -> NULL   -> row not admitted
```

It denies via NULL, not via FALSE. A reviewer checking for FALSE will find NULL and should not read it as an accident.

## Ledger

`apps/api/db/rls-cross-tenant-policies.txt` is unchanged by design. `scripts/check-rls-cross-tenant.sh:46-48` excludes RESTRICTIVE policies by construction and names `site_scope` as the reason. That same script states its own blind spot at line 79: it cannot see a missing or weakened restrictive policy, so it passing is not evidence this is right.

## This does NOT close the gap

Eight further site-keyed tables are owed policies — including `site_file_baseline` and `site_managed_files` ("changed core files") and `site_media_assets` / `media_optimization_jobs` / `site_media_settings` ("media library waste"), both named verbatim in m131's group descriptions, plus `site_security_bans`, `site_security_hardening_config`, and the indirect child `media_variant_results`.

They are deferred because **none of them exists in `db/schema.sql` at all**. `schema.sql` lags the migrations on whole tables, not only on policies. `CREATE POLICY` against an undefined relation would break sqlc generation, so the second increment is: backfill those eight `CREATE TABLE` definitions into `schema.sql`, then a new ordinal carrying their policies.

## Review focus

RLS and tenancy — needs `security-reviewer`. The specific things to attack: whether the single `FOR ALL` shape is right for the six tables here with a **nullable** `site_id` (DECISION 3 argues no table here has m112's org-inheritance shape), and whether any of the 22 has a reader that legitimately needs a NULL-`site_id` row.

### DIFF
--- a/apps/api/db/schema.sql
+++ b/apps/api/db/schema.sql
@@ -6059,3 +6059,434 @@ GRANT SELECT, INSERT, UPDATE, DELETE ON mcp_grants              TO wpmgr_app;
 GRANT SELECT, INSERT, UPDATE, DELETE ON mcp_connection_tokens   TO wpmgr_app;
 GRANT SELECT, INSERT, UPDATE, DELETE ON mcp_oauth_clients       TO wpmgr_app;
 GRANT SELECT, INSERT, UPDATE, DELETE ON mcp_authorization_codes TO wpmgr_app;
+
+-- ---------------------------------------------------------------------------
+-- m132 (ADR-061 A11, Wave 1.2): RESTRICTIVE app.site_scope policies on the
+-- tables the assistant reads.
+--
+-- Mirrors apps/api/migrations/20260903000000_m132_assistant_read_site_scope_rls.sql.
+-- That file is authoritative; this one is sqlc's input. Read its DECISION 2
+-- before touching the predicate: the nullif is the fail-closed property, not
+-- defensive noise. An empty allowlist denies via NULL, not via FALSE.
+--
+-- These are inert on every path that does not set app.site_scope, which today
+-- is every path except InScopedTenantTx (internal/db/db.go:582).
+-- ---------------------------------------------------------------------------
+
+CREATE POLICY cache_purge_audit_site_scope ON cache_purge_audit
+    AS RESTRICTIVE FOR ALL
+    USING (
+        coalesce(current_setting('app.site_scope', true), '') <> 'on'
+        OR site_id = ANY (
+            string_to_array(
+                nullif(current_setting('app.allowed_site_ids', true), ''), ','
+            )::uuid[]
+        )
+    )
+    WITH CHECK (
+        coalesce(current_setting('app.site_scope', true), '') <> 'on'
+        OR site_id = ANY (
+            string_to_array(
+                nullif(current_setting('app.allowed_site_ids', true), ''), ','
+            )::uuid[]
+        )
+    );
+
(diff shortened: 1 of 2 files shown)


mosamlife:
## Integration suite: one pre-existing failure, not from this branch

`make test-integration` from the repo root, real exit status, unpiped:

```
FAIL  github.com/mosamlife/wpmgr/apps/api/tests  895.023s
ok    .../tests/contract   3.268s
ok    .../tests/gh458      8.638s
ok    .../tests/testinfra  2.172s
EXIT=2
```

Exactly one test fails (`grep -c '^--- FAIL'` → 1):

```
--- FAIL: TestS7CapabilityOutsideTheOrgCeilingIsNotListedAsAppRole
  adr064_s7_mcp_tool_registry_rls_test.go:518: create mcp grant:
  ERROR: new row for relation "mcp_grants" violates check constraint
  "mcp_grants_site_scope_mode_check" (SQLSTATE 23514)
```

**It is not this branch.** Checked out the base commit `73d75494` (this branch's HEAD~1, no m132 present) and ran the single test:

```
scripts/with-machine-lock.sh api-integration -- sh -c 'cd apps/api && go test -count=1 \
  -run TestS7CapabilityOutsideTheOrgCeilingIsNotListedAsAppRole -timeout 10m ./tests/'
BASE EXIT=1
--- FAIL: TestS7CapabilityOutsideTheOrgCeilingIsNotListedAsAppRole
  ... violates check constraint "mcp_grants_site_scope_mode_check" (SQLSTATE 23514)
```

Identical failure, same constraint, without m132. The constraint is from m124; this branch touches neither `mcp_grants` nor any constraint.

### The part worth acting on separately

`git merge-base --is-ancestor 73d75494 origin/main` → **yes**. So the integration suite is currently **red on `origin/main`**. `ci.yml` does not run that package, which is how it merged green — the exact failure mode `.claude/rules/ci-and-build-logic.md` describes. Worth its own issue; it is not a blocker for this PR and this PR does not fix it.

PR #649, opened 2026-08-31
fix(mcp): route ResolveScopeSites on the principal, not a bare tenant id

mosamlife:
Closes ADR-061 A11 item 2 at `ResolveScopeSites`. Draft: needs a `security-reviewer` pass.

## What was wrong

`mcp.Repo.ResolveScopeSites` took a bare `tenantID uuid.UUID` and ran `InTenantTx`, so it could not route on scope even in principle. `app.site_scope` was never set, and the RESTRICTIVE `_site_scope` policies — `sites_site_scope` (m19) above all, the one this query's join through `sites` passes through — were switched off at the function whose job is to enforce scope. m132's 22 new policies inherit the same problem: they are inert unless `InScopedTenantTx` sets the GUC.

It now takes `db.ScopedPrincipal` and runs `RunTenantTx`.

## The two call sites are different problems

**`mint.go` `verifyScopeReferents` — threads the operator's principal.** Mode `'list'` verifies requested site ids by resolving them and checking each came back, so a site-scoped operator resolving a site outside their allowlist would be minting a credential wider than themselves. That path is **unreachable today**: `MintConnection` step 1 calls `requireOrgScopedPrincipal` (`internal/mcp/service.go:1648`), which refuses a site-constrained principal outright, so every principal arriving here is org-scoped and `RunTenantTx` dispatches it exactly as `InTenantTx` did. This changes no behaviour by one bit. It is the fail-closed backstop for the day that guard moves — the same reason `IsSiteConstrained` carries its second disjunct.

**`service.go` `Authenticate` — stays tenant-scoped, deliberately.** This is a bootstrap: it has no `domain.Principal` at all, and it resolves the grant's own `SiteScopeMode`/`ScopeTagIds`/`ScopeSiteIds` into concrete ids. That resolution is what *produces* the allowlist, so scoping it by the allowlist it is computing is circular:

- `'tags'` — the site set is not known until the query runs. There is no allowlist to supply; the empty one resolves every tag-scoped connection to zero sites.
- `'all'` — names no ids; org-wide is the stored grant's meaning.
- `'list'` — `sites_site_scope`'s predicate is `id = ANY(allowed)` and the query's own is `s.id = ANY($3)` over the same array, so the policy intersects the set with itself and can refuse nothing the query does not already refuse.

It now goes through a named `bootstrapTenantPrincipal`, so the exception is stated in code at its own call site instead of being indistinguishable from an oversight — which is the real win of the signature change.

**A11 item 2 as written is over-broad.** See the comment below.

## Proofs

New: `apps/api/tests/adr061_a11_scope_chokepoint_helper_test.go`. Through `mcp.Repo`, opening no connection and hand-setting no GUC, asserting `current_user`/`rolsuper`/`rolbypassrls` from inside a transaction the helper under test opened.

The policies engage — same query, same two sites, same tenant, differing only in the principal:

```
adr061_a11_scope_chokepoint_helper_test.go:88: connected as "wpmgr_app" rolsuper=false rolbypassrls=false
adr061_a11_scope_chokepoint_helper_test.go:114: org-scoped principal, mode 'list' naming 2 sites, resolved to 2
adr061_a11_scope_chokepoint_helper_test.go:129: site-scoped principal (allowlist of 1), same 2 ids, resolved to 1
```

Both sites are in the **same** tenant on purpose: a cross-tenant fixture would pass identically with the site-scope policy off.

Over-fire cases (org-scoped operator, unauthenticated bootstrap, mode `'all'`, and the empty allowlist meaning zero sites) are covered and green.

`go test ./tests/ -run 'TestMCP|TestA11|TestS16|TestS6b' -count=1` → `ok ... 71.762s`.

## Also owed, not done here

The m132 invariant test — "every site-keyed table carries a site-scope policy" — is its own job. It needs a catalogue sweep with a defensible definition of "site-keyed" and an explicit exemption list, and half-doing it would ship a guard that passes when it finds nothing. Briefing it separately.

### DIFF
--- a/apps/api/internal/mcp/mint.go
+++ b/apps/api/internal/mcp/mint.go
@@ -389,7 +389,7 @@ func (s *Service) MintConnection(ctx context.Context, req MintConnectionRequest)
 	// verifyScopeReferents -- and read its comment before adding any
 	// "resolves to no sites" refusal here, because that would be a defect
 	// rather than a hardening.
-	if err := s.verifyScopeReferents(ctx, req.Principal.TenantID, req.SiteScope); err != nil {
+	if err := s.verifyScopeReferents(ctx, req.Principal, req.SiteScope); err != nil {
 		return MintedConnection{}, err
 	}
 
@@ -588,7 +588,24 @@ func (s *Service) resolveMintCapabilities(requested []Capability) (CapabilitySet
 // Distinguishing them is what makes the empty-scope acceptance safe. Without
 // this check, "accept empty" would also silently accept a typo.
 //
-func (s *Service) verifyScopeReferents(ctx context.Context, tenantID uuid.UUID, req SiteScopeRequest) error {
+// IT TAKES THE WHOLE PRINCIPAL AND NOT A tenantID (ADR-061 A11 item 2). The
+// site-id arm below resolves ids through the audited chokepoint, and the
+// chokepoint routes on the principal's scope: a site-constrained operator must
+// not be able to name a site outside their own allowlist and have it verify,
+// because verifying is what lets the mint store it. That would be an operator
+// minting a credential wider than themselves.
+//
+// TODAY THAT PATH IS UNREACHABLE, AND THIS IS STILL NOT REDUNDANT.
+// MintConnection's step 1 calls requireOrgScopedPrincipal, which refuses a
+// site-constrained principal outright, so every principal arriving here is
+// org-scoped and RunTenantTx dispatches it exactly as InTenantTx did. Passing
+// the principal changes no behaviour by one bit today. It is the fail-closed
+// backstop for the day that guard moves or a site-scoped mint becomes a
+// product decision -- the same reason IsSiteConstrained carries its second
+// disjunct, and the same reason ADR-061 A11 lists three independent gates on
+// this path and says none of them is redundant.
+func (s *Service) verifyScopeReferents(ctx context.Context, principal domain.Principal, req SiteScopeRequest) error {
+	tenantID := principal.TenantID
 	switch req.Mode {
 	case SiteScopeModeTags:
 		// The tenant's tag registry, read under tenant RLS. Tags are a small,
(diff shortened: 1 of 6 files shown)


mosamlife:
Review findings addressed in `013b532b`. `origin/main` merged.

**1. `requireOrgScopedPrincipal`'s doc comment restored.** Both functions now carry their own; `mint.go`'s pointer at "requireOrgScopedPrincipal's own doc" resolves again.

**2. The "no behaviour by one bit" claim was false and is corrected in all three places** — commit message, `mint.go`, and the test comment that said "byte-for-byte". It now says what actually changes (`InTenantTxAsUser`, one extra `set_config('app.user_id', …)` per resolution) and why the resolved set cannot widen (`sites_shared_read` is PERMISSIVE FOR SELECT over *other* tenants' shares, excluded by `WHERE s.tenant_id = $1`).

**3. `scopeSitesScoped` is now asserted, not deleted.** The unit fake *can* distinguish the two cases — it sees the principal — so the recorder became `TestAuthenticateHandsTheChokepointAnUnscopedPrincipal`. It also checks the chokepoint was actually reached, so a `false` reading from a call that never happened cannot pass vacuously.

I chose this over the suggested call-site-count guard. It is cheaper *and* stricter: a count fires on a proxy and goes stale as call sites legitimately move, whereas this fails on the defect itself. **MUTATION M3** — `bootstrapTenantPrincipal` returning `Scope: domain.ScopeSite`, i.e. A11 item 2 applied literally:

```
--- FAIL: TestAuthenticateHandsTheChokepointAnUnscopedPrincipal (0.00s)
    oauth_test.go:1030: Authenticate handed the scope chokepoint a SITE-CONSTRAINED
    principal. That resolution is what PRODUCES the allowlist, so scoping it by the
    allowlist it is computing is circular ...
```

Restored, green, `git grep "MUTATION M" -- apps/api` exits 1 and the tree is clean. Same defect the reviewer's M2 caught through the e2e suite, now caught in 0.00s with no container.

**4. Stale prose:** fixed in `mint.go`. **`db/query/mcp_connections.sql:650` is NOT fixed and is not mine to fix** — `apps/api/db/query/**` is `database-engineer`'s path. The header still says "MUST RUN INSIDE InTenantTx"; it needs "inside a tenant transaction chosen by the principal's scope". Please route it.

**Both conditions recorded as conditions.** The `'list'` idempotence argument now states the two predicates it depends on and says nothing binds them together. `bootstrapTenantPrincipal`'s comment says outright that it is indistinguishable from any org-scoped no-user principal at a dispatch that reads three fields, and names the test that reddens.

Post-merge: `go vet ./internal/... ./tests/... ./cmd/...` clean; `go test ./tests/ -run 'TestA11|TestMCP' -count=1` → `ok ... 78.424s`.

Issue #647, opened 2026-08-31
A test that never passed merged to main: apps/api/tests is outside ci.yml


mosamlife:
## What happened

`TestS7CapabilityOutsideTheOrgCeilingIsNotListedAsAppRole` (`apps/api/tests/adr064_s7_mcp_tool_registry_rls_test.go:494`) is red on `origin/main` and has been since it was introduced in `49e8aff3`. It has never passed once.

The cause is a one-word test defect. `s7GrantWithBearer`'s fourth parameter is `mode string`, which goes to `SiteScopeMode`:

```
$ grep -n "s7GrantWithBearer(" apps/api/tests/adr064_s7_mcp_tool_registry_rls_test.go
170:	_, bearer := s7GrantWithBearer(t, mcpRepo, tenant,  "list", []uuid.UUID{s1.ID})
241:	_, bearer := s7GrantWithBearer(t, mcpRepo, tenant,  "list", []uuid.UUID{s1.ID})
309:	_, bearer := s7GrantWithBearer(t, mcpRepo, tenant,  "list", []uuid.UUID{s1.ID})
418:	_, bearer := s7GrantWithBearer(t, mcpRepo, tenantA, "list", []uuid.UUID{siteB.ID})
518:	_, bearer := s7GrantWithBearer(t, mcpRepo, tenant,  "ceiling", []uuid.UUID{s1.ID})
```

The test's own descriptive label landed in the mode argument. `mcp_grants_site_scope_mode_check` (m124, `apps/api/migrations/20260826000000_m124_mcp_connection_surface.sql:623`) restricts the column to `('all','tags','list')`, so the insert fails 23514 and the test dies in setup before reaching a single assertion.

Confirmed pre-existing rather than caused by any open branch: `git merge-base --is-ancestor 73d75494 origin/main` returns true, and the same failure reproduces at that commit with no recent migration present.

The one-word fix is tracked separately. **This issue is about why nobody knew.**

## Why nobody knew

`ci.yml` does not run `apps/api/tests`. That is a deliberate, documented decision — `.claude/rules/ci-and-build-logic.md` names the package and gives the reason, and the working agreements say to run `make test-integration` locally before merging anything touching RLS, tenant scoping, the email domain, or object-storage reclamation.

[...]

PR #661, opened 2026-09-01
release: 0.61.151

mosamlife:
Control-plane only. The agent plugin did not change, so every agent version surface stays at 0.61.147 and the marketing hero badge does not move.

## What is in it

Five migrations, m128 to m132. The two an operator needs to know about:

- **m132 adds 22 restrictive site-scope policies across 22 tables.** Every table the assistant surface can read now carries the same database-level site-scope opinion the rest of the schema already enforced.
- **A capability refusal answers 403 instead of 401.** This is a client-visible behaviour change. Anything treating 401 as "renew and retry" against the assistant surface must treat 403 as terminal, because re-authenticating could never change a refusal that is about the grant rather than the credential.

The rest is in the changelog entry.

## Verification

A full integration run against `4da49240`, under the machine lock:

```
ok  github.com/mosamlife/wpmgr/apps/api/tests            887.247s
ok  github.com/mosamlife/wpmgr/apps/api/tests/contract     2.836s
ok  github.com/mosamlife/wpmgr/apps/api/tests/gh458        7.669s
ok  github.com/mosamlife/wpmgr/apps/api/tests/testinfra    3.112s
```

0 occurrences of FAIL, no panic, no `[no tests to run]`, no build failure. That package is outside `ci.yml`, so this is the run that covers the RLS and tenancy proofs those five migrations touch.

The diff from `4da49240` to the release base is comment-only: 0 non-comment changed lines in both `apps/api/internal/db/sqlc/` and `apps/api/db/query/`, so the baseline holds for this release.

Worth noting for the record: that package was red before this batch. `TestS7CapabilityOutsideTheOrgCeilingIsNotListedAsAppRole` had never once passed since it was written, and `ci.yml` does not run the package, so nobody knew. Fixed in #648, cause written up in #647. This is the first clean run of it.

## Not done here, deliberately

**No tag.** The tag is the last step, after deploy and after QA.

## Summary by CodeRabbit

- **New Features**
  - Added a tenant-level switch to disable the assistant surface.
  - Expanded read capabilities to the complete v1 vocabulary.
  - Added per-connection and per-tenant MCP tool-call rate limits.
  - Connections now record their configured client.
  - Grant expiry is limited to a maximum of one year.

- **Bug Fixes**
  - Capability refusals now return HTTP 403 with specific error details.

- **Security**
  - Added site-scope protections across supported data areas.
  - Improved site resolution using the authenticated caller.

- **Documentation**
  - Updated release, installation, image, and API version references to 0.61.151.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,8 +22,8 @@ House rules: no em dashes, no en dashes, no competitor names. Use "to" for range
 
 ### Security
 
-- Every table the assistant surface can read now carries the same restrictive site-scope policy already enforced elsewhere in the schema: 22 policies added across 22 tables that previously had no database-level opinion of their own about which site a scoped grant could reach (m132).
-- The chokepoint the assistant surface uses to resolve which sites a connection may reach now routes on the caller's authenticated principal rather than a bare tenant id, so the site-scope policies above actually engage instead of being satisfied by any grant in the tenant.
+- Every table the assistant surface can read now carries the same restrictive site-scope policy already enforced elsewhere in the schema: 22 policies added across 22 tables that previously had no database-level opinion of their own about which site a scoped grant could reach (m132). These protect every path that runs with a site-constrained principal, which is the ordinary way a site-scoped collaborator reaches this data.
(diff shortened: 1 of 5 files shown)


mosamlife:
Fixed. Verified independently: `bootstrapTenantPrincipal` (apps/api/internal/mcp/service.go:1678) returns only TenantID, and `Authenticate`'s call to `ResolveScopeSites` (service.go:1143) uses it, so `dispatchTenantTx` (apps/api/internal/db/db.go) never routes that call through `InScopedTenantTx` on the live path. The mint path is likewise guaranteed non-site-constrained by `requireOrgScopedPrincipal`. Narrowed both CHANGELOG.md and the marketing changelog page to say the m132 policies protect the ordinary site-scoped product surface, and that the chokepoint is now capable of engaging them but does not on the assistant's own read path, which relies on the resolved allowlist in application code instead. Pushed as 326b0b15.

PR #662, opened 2026-09-01
fix(mcp): audit is fail-closed on the assistant surface, reads included

mosamlife:
## The finding that matters most: a real audit failure did not look like a fake one

The Go-layer proofs for this change passed from the first commit. They were not sufficient, and the integration proof — a real `audit_log` append refused for a real reason, as `wpmgr_app` — went red immediately and found a live defect:

```
append-refused response: HTTP 200 body={"jsonrpc":"2.0","id":7,"error":{"code":-32602,"message":"failed to append audit entry"}}
```

Two defects in one response. **-32602 is `invalid params`** — the caller told their request was malformed when the fault was entirely server-side — and **the message names the failure**, handing any caller the audit-liveness oracle in plain English rather than merely inferable from a code.

**Why every unit test passed throughout, which is the transferable part.** `internal/audit` returns *typed* domain errors: `audit_insert_failed`, `audit_prev_failed`, `audit_marshal_failed`, `audit_canonical_failed`. The wire mapping in `toolError` only recognised `mcp_audit_unavailable`, the code produced by the nil-recorder guard, so a genuine append failure arrived as a domain error with a code it had never heard of and fell through to the default branch. The fake recorder in the unit tests returns `errors.New`, which is **not** a domain error, so it took the non-domain branch and produced the correct generic answer. The fake could not express the bug that the real database produced on its first run.

Fixed by `auditFailure` (`internal/mcp/service.go`), which normalises anything the audit package returns into `ErrCodeAuditUnavailable` while preserving the original as the cause. All three record sites route through it, so a new error code added inside `internal/audit` later cannot reintroduce the fallthrough.

## What the branch does

Makes the assistant surface's audit **fail-closed, reads included**, per ADR-061 A10 and the owner ruling. The surface previously did the opposite for reads and argued for it in a doc comment on `RecordToolCall`.

**1. The live-surface records fail closed.** `RecordToolCall`, `RecordToolDenied` and `RecordProtocolDenied` propagate their append failure; `callTool` and `writeProtocolRefusal` withhold what they were about to send.

The append sits **after the read, before the answer is serialised**:

- A read writes no row, so a shared transaction has nothing to roll back. All it would add over this ordering is that the append commits before the answer — which the ordering already gives — at the cost of threading a `pgx.Tx` through `Store` and every tool implementation.
- The action being recorded is the **disclosure**. Data that never leaves the process was never seen, so binding the record to the disclosure is the faithful reading of "no AI action is ever performed whose record was lost".
- Recording *before* invoke was rejected: it writes `mcp.tool.called` rows for calls that then failed and answered nothing, and an auditor cannot distinguish those from real ones.

This stops being true the moment a **write** tool lands — that append belongs in the write's own transaction via `RecordInTx`, because then there is a row to roll back and the ordering trick no longer substitutes for atomicity. Said so, with the reason, in the code.

**2. The fail-open helper is unreachable from `internal/mcp` — as a compile error.** `audit.Recorder` gains `RecordOrFail` (same mechanics as `Record`, opposite contract: the caller propagates). The package's `auditRecorder` interface lists `RecordOrFail` and `RecordInTx` and **not** `Record`. A grep guard would catch the same mistake one CI run later and only for anticipated spellings; a method set catches it in the editor, for every spelling.

**3. Approve, mint and revoke fail closed too.** A10 names approval. `Approve` (`service.go`), `MintConnection` (`mint.go`) and `RevokeConnection` each returned `nil` on a missing recorder — creating the grant, or minting the credential, and skipping the record. Those are the most consequential rows this system writes: an operator granting an AI assistant access to their fleet with no record of it. The error now propagates out of the callback into the caller's transaction, which rolls the write back, so there is no half-state where the credential exists and the row explaining it does not.

Revocation is closed alongside them although A10 does not spell it out. Leaving one sibling open is how a deliberate posture gets read as an inconsistency and tidied back to `return nil` later, and a revocation is the row an operator most needs after an incident.

**4. A recorder-less `Service` refuses instead of serving.** Documented as supported ("simply does not audit"); under the ruling it is the one state that must not exist — an unaudited assistant surface presenting as a working one.

**5. The batching rule, written down** (on the `Service.audit` field): reads are **not batched and not sampled** — one row per `tools/call`, synchronously, ahead of the answer. The volume assumption is the per-tenant tool-call ceiling the limiter already enforces; retention is `audit_log`'s own. Sampling would have to be re-argued if that ceiling rises, because a sampled read log can only answer "some of them".

## Executed evidence

Integration, `./scripts/with-machine-lock.sh wpmgr-integration -- go test ./apps/api/tests/ -run TestMCPAuditFailClosed -count=1 -v`:

```
connected as "wpmgr_app" rolsuper=false rolbypassrls=false
audit_log INSERT for "wpmgr_app" is now false (verified with has_table_privilege)
append-refused response: HTTP 200 body={"jsonrpc":"2.0","id":7,"error":{"code":-32603,"message":"the tool call failed"}}
audit_log INSERT for "wpmgr_app" is now true (verified with has_table_privilege)
GREEN: answer served, exactly 1 mcp.tool.called row, actor=assistant/272dbbe5-6ca1-4cd6-b400-1dec6b09fe6f
--- PASS: TestMCPAuditFailClosed_NoAnswerIsServedWhenTheAppendCannotCommit (9.24s)
--- PASS: TestMCPAuditFailClosed_RefusalAndInternalErrorAreByteIdentical (7.06s)
ok  github.com/mosamlife/wpmgr/apps/api/tests  18.720s
```

The byte-identical pair, from the same run:

```
(a) audit unavailable : HTTP 200 body={"jsonrpc":"2.0","id":7,"error":{"code":-32603,"message":"the tool call failed"}}
(b) generic internal  : HTTP 200 body={"jsonrpc":"2.0","id":7,"error":{"code":-32603,"message":"the tool call failed"}}
```

**How the failure is provoked, and why not a closed pool.** The superuser pool runs `REVOKE INSERT ON audit_log FROM wpmgr_app`. The connection stays healthy, the tenant's rows stay readable, the tool's own read still succeeds, and the only thing that breaks is the append. A closed pool would break the read too — and then the test cannot distinguish "the answer was withheld" from "there was no answer to withhold". The privilege is read back with `has_table_privilege` rather than trusted, and the row is counted with a query rather than inferred from the response. The role is asserted and printed from `pg_roles` **inside a transaction the request path itself opens**, because a `REVOKE` against `wpmgr_app` is inert if the code under test connects as a superuser or `BYPASSRLS` role, and the whole file would then pass by answering both halves successfully.

**A correction worth recording.** The first version of the byte-identical test provoked its comparison case by revoking `SELECT ON sites`. That looked like an unrelated internal failure but actually broke *authentication*, answering `HTTP 500 "the request could not be authenticated"` from a different handler — two responses that were never comparable, which would have read as a pass. It now compares against the literal envelope `toolError` emits for any non-domain infra failure, the branch every other server-side tool fault lands on.

Unit-level mutations, each red then green on restore:

- **M1**, fail-open answer restored in `callTool` → red, and the failure output was the full site payload leaking into the response.
- **M2**, `requireRecorder`'s guard removed → red.
- **M3**, `s.audit.Record(...)` reintroduced → `internal/mcp/service.go:1487:20: s.audit.Record undefined (type auditRecorder has no field or method Record)`.

The new tests assert on the **absence of the site payload from the response bytes**, not on the JSON-RPC code — a code-only assertion would go green on a surface that leaks the answer inside the failure.

## Tests changed, and why that is not a weakened guard

14 unit tests broke on the recorder-less guard. They built Services without a recorder — a configuration production has never been in, since `main.go` has always called `WithAudit` — so they were testing a shape that never ships. They now construct the Service the way production does, via an `auditedService` helper. **The guard was not relaxed to accommodate them.** The one test that is genuinely *about* the unaudited case still calls `NewService` directly and asserts the refusal.

`TestTransport_AuditGapLogsReasonAndPhaseSeparately` had its central assertion inverted: a protocol refusal whose row cannot be written is now `500`, not the typed `400` naming the floor and target. It also now asserts neither half of the refusal survives into the internal error.

## The residual, and what review should focus on

**A denied caller can distinguish "refused" from "refused and unrecordable."** That is a real information leak and the one point of the old `auditGap` argument that survives: the caller on that branch is by definition someone probing for something they may not have, which is what `AuthorizeTool`'s whole disclosure decision exists to defeat.

It is narrowed — an audit failure and any other server-side failure are now byte-identical, proven above — and then **accepted**, against the alternative of a surface that keeps answering while recording nothing. It is not mitigated away.

The two things worth the reviewer's attention: the **`auditFailure` mapping** (it is now the single chokepoint for every audit error reaching the wire, so a gap there reopens the oracle) and the **residual itself**.

`auditGap` is kept and re-documented: with the caller now getting an error carrying none of the lost row's facts, that log line is the only place they exist.

## Summary by CodeRabbit

* **Bug Fixes**
  * MCP requests now fail safely when audit records cannot be written, returning a generic internal error instead of serving responses without an audit trail.
  * Tool answers and detailed protocol refusals are withheld when auditing fails, preventing sensitive error details from being exposed.
  * Connection token minting no longer succeeds without a completed audit record.
* **Reliability**
  * Audit entries are confirmed committed before dependent responses are returned.

### DIFF
--- a/apps/api/internal/audit/audit.go
+++ b/apps/api/internal/audit/audit.go
@@ -769,6 +769,34 @@ func (r *Recorder) Record(ctx context.Context, e Event) (Entry, error) {
 	return out, err
 }
 
+// RecordOrFail appends an audit entry in its own transaction and returns only
+// once that entry is COMMITTED. Mechanically it is Record; the difference is
+// the contract, and the contract is the point.
+//
+// Record's doc tells its callers to log and continue. That instruction is
+// correct for the rest of the product and fatal for the assistant surface, and
+// a posture that exists only as a sentence in a doc comment is not a posture --
+// it is a suggestion the next caller reads as optional. RecordOrFail carries
+// the opposite instruction, and exists so that a package which must never fail
+// open can hold a field type whose method set DOES NOT CONTAIN Record. See
+// internal/mcp's auditRecorder: reaching the fail-open helper from there is a
+// compile error, not a review comment.
+//
+// THE CALLER MUST PROPAGATE THE ERROR, and "propagate" means the operation the
+// entry describes does not reach the client. A caller that logs this error and
+// answers anyway has the fail-open behaviour back with a fail-closed name on
+// it, which is worse than Record because the name now vouches for it.
+//
+// USE RecordInTx INSTEAD WHENEVER THERE IS A COMPANION WRITE. This helper
+// commits its own transaction, so it cannot roll anything back; it is the right
+// shape only where the effect being recorded is a DISCLOSURE rather than a row
+// (a read answered to a client, a refusal returned to a caller), and the caller
+// withholds that disclosure when this returns non-nil. The moment the recorded
+// operation writes a row, the append belongs in that row's transaction.
+func (r *Recorder) RecordOrFail(ctx context.Context, e Event) (Entry, error) {
+	return r.Record(ctx, e)
+}
+
 // RecordInTx appends an audit entry inside the CALLER's already-open
 // transaction, instead of opening its own. It is the fail-closed counterpart
 // to Record, required by ADR-064 Decision 7 (and, on paper, ADR-061 Decision
--- a/apps/api/internal/mcp/approve_test.go
+++ b/apps/api/internal/mcp/approve_test.go
@@ -130,7 +130,7 @@ func TestApprove_RefusesAClientThatWasNeverRegistered(t *testing.T) {
 // path, so the fix cannot regress into comparing the body against itself.
 func TestApprove_ResolvesTheClientBeforeMinting(t *testing.T) {
 	store := approvalStore()
-	if _, err := NewService(store).Approve(context.Background(), validApproval()); err != nil {
+	if _, err := auditedService(store).Approve(context.Background(), validApproval()); err != nil {
 		t.Fatalf("a legitimate approval was refused: %v", err)
 	}
 
@@ -246,7 +246,7 @@ func TestApprove_RefusesAnIncoherentOrEmptySiteScope(t *testing.T) {
 // RESOLVED client id rather than the body's copy.
 func TestApprove_ValidApprovalMintsAndRecordsTheResolvedClient(t *testing.T) {
 	store := approvalStore()
-	got, err := NewService(store).Approve(context.Background(), validApproval())
+	got, err := auditedService(store).Approve(context.Background(), validApproval())
 	if err != nil {
 		t.Fatalf("a legitimate approval was refused: %v", err)
 	}
@@ -384,7 +384,7 @@ func TestApprove_RefusesASiteScopedApprover(t *testing.T) {
 // policy; it can prove the principal arrived.
 func TestApprove_HandsTheWholePrincipalToTheStore(t *testing.T) {
 	store := approvalStore()
-	svc := NewService(store)
+	svc := auditedService(store)
 
 	req := validApproval()
 	if _, err := svc.Approve(context.Background(), req); err != nil {
--- a/apps/api/internal/mcp/audit_fail_closed_test.go
+++ b/apps/api/internal/mcp/audit_fail_closed_test.go
@@ -1,0 +1,192 @@
+package mcp
+
+import (
+	"errors"
+	"log/slog"
+	"net/http"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/gin-gonic/gin"
+	"github.com/google/uuid"
+
+	"github.com/mosamlife/wpmgr/apps/api/internal/audit"
+	"github.com/mosamlife/wpmgr/apps/api/internal/db/sqlc"
+)
+
+// ---------------------------------------------------------------------------
+// ADR-061 A10, fail-closed: no tool answer is served whose audit record was
+// lost.
+//
+// WHAT THESE PIN, AND WHY THE ASSERTION IS ABOUT THE BODY RATHER THAN THE CODE.
+// The property is not "an unrecordable call returns an error" -- that would
+// still hold if the handler serialised the tool's text and then bolted an error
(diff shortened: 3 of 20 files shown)


mosamlife:
Verified against the code rather than taken, and declining the header comparison deliberately.

**Headers do not vary by error on this path.** Every `c.Header(...)` in `transport.go` belongs to a different path than the tool-call error:

```
$ grep -n "c.Header(" apps/api/internal/mcp/transport.go
335,336,356,357,361  CORS - identical on every response
376                  Allow          - the 405 path only
810                  Retry-After    - the rate-limit path only
1273                 WWW-Authenticate - the auth path only
```

A tool error is `200` plus a JSON-RPC body with no error-specific header. So comparing headers between an audit failure and a generic internal failure would add an assertion that cannot fail. This repository has been bitten repeatedly by exactly that shape, and its review guidance names it: a negative control that cannot fail is worse than none, because it reads as coverage.

The prose was the real defect and it is fixed — the comment previously claimed "byte for byte including the headers", which was untrue. It now states plainly that a header-only difference is outside what the test proves.

**On `genericInternal` being a hand-written literal** — that is a fair observation and it is answered by cross-reference rather than by duplication. `TestToolCall_AnswerIsWithheldWhenTheAuditAppendFails` (`apps/api/internal/mcp/audit_fail_closed_test.go:87`) drives a genuinely non-domain error through the same `toolError` call and asserts `codeInternalError`, so the literal is anchored by an executed test rather than by this file's say-so. Provoking a second unrelated internal failure at this exact point needs a fault that survives authentication and authorization while still breaking invoke; the first attempt at that produced an authentication failure instead and compared two different handlers, which is how the earlier version of this test was wrong.

Resolving. The substantive half of the finding was real and is fixed; the header half is a check that could not fail.

Issue #660, opened 2026-08-31
No screen lets an operator choose a connection's capabilities


mosamlife:
## The gap

The capability vocabulary now holds eight strings and seven are conferrable, so a connection can hold any of 2⁷−1 distinct capability sets. **There is no screen on which an operator can choose one.**

Tracing the three surfaces:

**The connect wizard** (`apps/web/src/features/ai-connections/connect-wizard.tsx`) has four steps — "Pick your client", "How it signs in", "Sites it may reach", "Set it up". Step 4 is the setup snippet, not capabilities. The file's own comment explains why it could never be: *"The grant is created by the approval screen at /connect/ai, which the CLIENT redirects into; this wizard never calls it and has no channel to hand it an answer."* The wizard produces configuration text. It creates nothing.

**The consent screen** (`apps/web/src/routes/_authed/connect.ai.tsx`) is where the grant is actually created, via `Approve` (`apps/api/internal/mcp/service.go:702`). It sends no capability field and there is none to send — `Approve` calls `DefaultGrantCapabilities()` unconditionally, which returns exactly `["mcp.sites.read"]`.

**The mint endpoint** is the only path that accepts a capability list. `MintConnectionRequest.Capabilities` (`apps/api/internal/mcp/dto.go:264`) is optional, narrowed against the ceiling, and refuses the whole request rather than trimming when a capability exceeds it. This is an API surface, not a screen.

So every connection an operator creates through the product holds one capability, and the other six conferrable ones are reachable only by calling the mint API directly.

## Why this is a design question, not a task

The obvious reading — "add Step 4 to the wizard" — cannot work. The wizard has no channel to the grant. Putting a picker there would collect an answer with nowhere to send it, which is worse than not asking.

That leaves real options, and they are not equivalent:

1. **Add capabilities to the consent screen.** It is the screen that creates the grant, and consent is the moment an operator is already deciding what to allow — which is exactly when a permission choice belongs. Cost: `Approve` needs the field, so this is a backend change before it is a frontend one, and the consent screen is reached by a client redirect, so its layout is constrained by what the OAuth flow can carry.
2. **Edit capabilities after creation, from the connections list.** Sidesteps the OAuth flow entirely and pairs naturally with showing what a connection currently holds. Cost: there is a window where the connection holds the preset, and "grant then narrow" is the wrong default direction for a permission system.
[...]

PR #664, opened 2026-09-01
fix(ai-connections): correct stale capability-schema/endpoint claim in connect wizard

mosamlife:
## Summary
- The connect wizard's comments claimed the capability vocabulary and the mint endpoint "do not exist yet." Both exist now: `apps/api/internal/mcp/policy.go` seats an eight-member capability vocabulary (seven conferrable via the read scope), and `apps/api/internal/mcp/dto.go` carries `capabilities` on both the mint request (line 264) and the mint response (line 305).
- The same comment mislabelled the wizard's fourth step as "capabilities"; `STEPS` actually names it "Set it up" (the setup snippet). There is no capabilities step.
- The frontend genuinely has nothing generated to bind a picker to (`packages/openapi-client/src/generated`'s only `capabilities` type is the unrelated `ObjectCacheCapabilities`), but that's not why a picker isn't rendered here: this wizard never creates the grant, so a picker here would collect an answer with nowhere to send. Reworded both comments to state that, and referenced GH #660 (open) which tracks where such a picker should actually live.
- The test that pinned the old, now-false claim ("does not invent a numbered step for capabilities, which has no backend") asserted a real, still-true structural property (no capabilities heading; the rail says permissions are chosen on the approval screen) under a false title/comment about the backend. Renamed and re-commented it to assert only the frontend-keepable fact.

## Verification
- `pnpm -C apps/web typecheck` — clean.
- `pnpm -C apps/web lint` — 0 errors, 10 pre-existing warnings unrelated to this change.
- `pnpm -C apps/web test` — 166 files / 2110 tests passed.
- `apps/web/node_modules/.bin/impeccable detect apps/web/src/features/ai-connections` — exit 0, no findings.
- Mutation-tested the renamed assertion: broke the rail's "Permissions are chosen on the approval screen" text, `npx vitest run src/features/ai-connections/connect-wizard.test.tsx -t "does not invent a numbered step"` went red naming that exact line; restored, it went green again.

## Test plan
- [x] CI: typecheck, lint, unit tests
- [ ] Reviewer confirms the corrected comment text against `apps/api/internal/mcp/policy.go` and `dto.go:264`/`dto.go:305`

## Summary by CodeRabbit

* **Documentation**
  * Clarified that the AI connections wizard does not currently provide a step for configuring capabilities.
  * Updated related test descriptions and comments to accurately reflect the available capability support and wizard behavior.

### DIFF
--- a/apps/web/src/features/ai-connections/connect-wizard.test.tsx
+++ b/apps/web/src/features/ai-connections/connect-wizard.test.tsx
@@ -306,7 +306,11 @@ describe("step 3 exists at all, and sits before capabilities", () => {
     expect(screen.queryByText(/4\. Choose sites and permissions/i)).not.toBeInTheDocument();
   });
 
-  it("does not invent a numbered step for capabilities, which has no backend", async () => {
+  it("does not invent a numbered step for capabilities, which this wizard has no channel to write", async () => {
+    // Not a claim about the backend (the capability vocabulary and the mint
+    // endpoint both exist, GH #660) -- a claim about this file: it never
+    // creates the grant, so a step here would collect an answer with nowhere
+    // to send it, and the rail names where that answer is actually written.
     await reachSiteStep();
     expect(screen.queryByRole("heading", { name: /what this connection may do/i })).toBeNull();
     expect(screen.getByText(/permissions are chosen on the approval screen/i)).toBeInTheDocument();
--- a/apps/web/src/features/ai-connections/connect-wizard.test.tsx
+++ b/apps/web/src/features/ai-connections/connect-wizard.test.tsx
@@ -306,11 +306,16 @@ describe("step 3 exists at all, and sits before capabilities", () => {
     expect(screen.queryByText(/4\. Choose sites and permissions/i)).not.toBeInTheDocument();
   });
 
-  it("does not invent a numbered step for capabilities, which this wizard has no channel to write", async () => {
-    // Not a claim about the backend (the capability vocabulary and the mint
-    // endpoint both exist, GH #660) -- a claim about this file: it never
-    // creates the grant, so a step here would collect an answer with nowhere
-    // to send it, and the rail names where that answer is actually written.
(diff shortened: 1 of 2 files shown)


mosamlife:
Confirmed and fixed in f40a0c0e. The prior correction's "this wizard never creates the grant" was itself false, exactly as flagged:

- OAuth path: `Approve` (`apps/api/internal/mcp/service.go:607`) takes an `ApprovalRequest{Principal, Consent, GrantName, SiteScope}` and nothing else, so this wizard genuinely has no channel there.
- Token path: `TokenMintPanel` calls `useMintConnection()` -> `POST /api/v1/mcp/connections`, and the mint endpoint already accepts `capabilities` on both the request (`dto.go:264`) and response (`dto.go:305`). The only gap is a missing field on the frontend's `MintConnectionInput` (`use-ai-connections.ts:228-233`, which has `name`/`siteScopeMode`/`scopeTagIds`/`scopeSiteIds` and no `capabilities`).

Reworded both comments to name the two paths separately with file:line citations instead of a summarized claim, per the "why these keep going stale" ask. Did not add a picker to either path -- that's still open per #660.

Checked the renamed test: `reachSiteStep()` clicks the oauth auth card, so the assertion runs against the OAuth path specifically, where the claim is actually true. Updated the test's comment to say so and to note the token-path nuance, without changing what it asserts (it's a real, still-true DOM property for both paths: neither renders a capabilities picker today). Mutation-tested again: broke `<span>Permissions are chosen on the approval screen</span>` -> red, naming `connect-wizard.test.tsx:321`; restored -> green (`npx vitest run src/features/ai-connections/connect-wizard.test.tsx -t "does not invent a numbered step"`).

Gate re-run clean: `pnpm -C apps/web typecheck` clean; `pnpm -C apps/web lint` 0 errors/10 pre-existing warnings; `pnpm -C apps/web test` 166 files / 2110 tests passed; `impeccable detect apps/web/src/features/ai-connections` exit 0.

Issue #652, opened 2026-08-31
An operator cannot see what capabilities a live connection holds


mosamlife:
## What an operator cannot see

`connectionFromGrant` (`apps/api/internal/mcp/service.go:1783`) does not populate the capability set, so the connections list renders none. There is no surface anywhere — UI or API — that answers "what can this connection actually reach?"

Until now that cost nothing. Every grant held exactly one capability, `mcp.sites.read`, so an omitted field was indistinguishable from a correctly populated one; there was only ever one answer.

That changes with the capability-vocabulary widening. Seven capabilities become conferrable, so 2⁷−1 distinct sets are mintable through the endpoint. The fact becomes variable in the same release that removes the only reason it was safe to omit.

## Why this is a bug and not tech debt

An operator holds credentials that grant an AI assistant access to their WordPress fleet. The question "what is this credential allowed to do" is the single most important thing about it, and right now the honest answer is: read the database.

Concretely, an operator cannot:

- audit what they granted, after granting it
- notice a credential that holds more than they intended
- decide whether a connection is safe to leave live
- confirm that revoking or re-minting actually narrowed anything

A customer can feel every one of those, which is what separates this from debt with no user-visible symptom.

## What it needs

Populate the capability set in `connectionFromGrant`, and surface it wherever a connection is listed or inspected. The stored column is already the authority — `Authenticate` reads `chk.GrantCapabilities` off the same row and the same transaction as the `authorized` verdict — so this is a display gap, not a modelling one.

Two things worth getting right rather than shipping quickly:

1. **Show the stored set, not a recomputed one.** Recomputing from the scope registry is a guess that agrees with the row only until they diverge, and `service.go` already carries a comment explaining that this precise mistake was made once before and fixed by m127. The list must read the column.
2. **Show what the capability *means*, not the wire string.** `mcp.sites.read` is a value in a closed vocabulary, not a label for a human. The consent screen already has to solve this problem; whatever vocabulary it uses should be the same one here, or an operator will be comparing two different descriptions of the same grant.

## Not established

Whether the connections list is the only surface with this gap, or whether the same omission exists on the single-connection detail view and in any API response a customer could script against. Nobody has swept the read paths.

Whether the consent screen and this list would agree on wording today — they were written at different times against a one-capability world, and neither has been checked against the other since the vocabulary widened.

mosamlife:
Fixed and shipped. Verified against `main` at `bbd1f727`, deployed as 0.61.153.

[...]

Issue #653, opened 2026-08-31
A scope can confer a capability outside the vocabulary, and every test stays green


mosamlife:
## The gap

The capability work pins two closed sets against each other: `capabilityVocabulary` (what may be spelled) against the database CHECK, and `scopeCapabilities` (what a scope confers) against an explicit literal. Both directions of the vocabulary/CHECK pair are covered.

One direction of the *other* pair is not. A name added to `scopeCapabilities` but **not** to `capabilityVocabulary` is silently dropped by `NewCapabilitySet`, and `TestScopeReadConfersTheSevenReachableGroups` still passes — it compares `OrgDefaultCapabilities`' output against a seven-member literal, and the unknown name has already been discarded by the time the comparison happens.

So a scope can confer strictly less than its own map says, and every test stays green.

## Why it is low severity

The drop is **fail-closed**: it removes authority, never adds it. Nobody gets access they were not granted; a capability the map claims to confer simply does not work, and the failure surfaces as a tool being unavailable rather than as a leak. `policy.go` argues explicitly that dropping is safe here for that reason, and that argument is correct.

It is worth fixing anyway because the symptom is confusing in the wrong direction: the map is the thing an engineer reads to answer "what does this scope confer", and it can be wrong while the tests that exist to pin exactly that question pass.

## The fix

A test asserting every value in `scopeCapabilities` is present in `capabilityVocabulary`, evaluated **before** any set is constructed — the construction is what hides the discrepancy. No database needed; it belongs beside the existing unit tests.

The test must go red when it finds a mismatch and must not pass vacuously if either map is empty. Plant a name in one and not the other, watch it redden, restore.

## Related

The same shape appeared in the vocabulary/CHECK pair and was closed by pinning both directions. This is the equivalent for the scope map, and it is the one direction that pairing does not reach.

PR #665, opened 2026-09-01
web: render AI connection capabilities on the connections list

mosamlife:
## Summary
- Closes the web half of #652 (server half merged in #663). `connectionSchema` in `apps/web/src/features/ai-connections/use-ai-connections.ts` was missing `capabilities`, so zod's plain `z.object` silently stripped the field the server always sends and nothing rendered it.
- Adds `capabilities: readonly string[]` to `AiConnection`, requires the field in the wire schema (the server never omits it and never sends `null` — `apps/api/internal/mcp/dto.go`'s `capabilityNames` always builds with `make([]string, 0, len(...))`), and renders a new "Capabilities" column in `connections-list.tsx`.
- An empty set renders as a distinct `No capabilities - inert` badge instead of a placeholder that could read as missing data, since `Authenticate` refuses a connection holding none outright (`apps/api/internal/mcp/policy.go`).
- New `apps/web/src/features/ai-connections/capabilities.ts` holds the label map for the eight `mcp.<noun>.read` wire strings (`apps/api/internal/mcp/policy.go`'s `capabilityVocabulary`), so the future capability picker (#660) reuses it instead of growing a second map. An unrecognised capability string still renders as its raw wire value, matching `capabilitiesFromColumn`'s "do not filter the column read back" behaviour on the server.

## Mutation proof (the specific defect being fixed)
Removed `capabilities` from `connectionSchema`, ran the suite: 4 tests went red, each naming the missing field (`expected undefined to deeply equal [...]` / `expected [Function] to throw an error`). Restored the field: all tests green again. Full output pasted in the session log.

## Test plan
- [x] `pnpm -C apps/web typecheck`
- [x] `pnpm -C apps/web lint` (0 errors; pre-existing unrelated warnings only)
- [x] `pnpm -C apps/web test` (full suite green; one file-level flake under parallel load in `connect-wizard.test.tsx`, unrelated to this change, reruns green in isolation)
- [x] `pnpm -C apps/web build` (routeTree.gen.ts unchanged, no new routes)
- [x] `apps/web/node_modules/.bin/impeccable detect apps/web/src/features/ai-connections apps/web/src/routes/_authed/ai` (exit 0, no findings)
- Assertions are on the exact capability set (not non-emptiness), on `[]` explicitly, and on an unrecognised capability string rendering as itself.

### DIFF
--- a/apps/web/src/features/ai-connections/capabilities.test.ts
+++ b/apps/web/src/features/ai-connections/capabilities.test.ts
@@ -1,0 +1,30 @@
+import { describe, it, expect } from "vitest";
+
+import { capabilityLabel, KNOWN_CAPABILITIES } from "./capabilities";
+
+// KNOWN_CAPABILITIES is DERIVED from CAPABILITY_LABELS's keys (see the file
+// header), so there is no longer a second list a mutation could add a name to
+// without also adding it here -- that failure mode is structurally impossible
+// rather than merely tested for, and a test asserting the two agree would be
+// tautological (they are the same object read two ways). What is left to test
+// is the behaviour capabilityLabel actually promises.
+
+describe("capabilityLabel", () => {
+  it("labels every name in the known vocabulary", () => {
+    // Not hardcoding the set: walks whatever KNOWN_CAPABILITIES actually is.
+    for (const cap of KNOWN_CAPABILITIES) {
+      const label = capabilityLabel(cap);
+      expect(label.length).toBeGreaterThan(0);
+      // A known capability gets an actual label, not its own wire string back
+      // -- that fallback is reserved for names outside the vocabulary.
+      expect(label).not.toBe(cap);
+    }
+  });
+
+  it("falls back to the raw wire string for a capability this build does not know", () => {
+    // The property #652 was filed over, preserved: an unrecognised name the
+    // server actually stored still renders, as itself, rather than being
+    // dropped or replaced with a generic "unknown" placeholder.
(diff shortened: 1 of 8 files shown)


mosamlife:
Fixed in 5fc09f8a. Verified the Greptile finding: KNOWN_CAPABILITIES and CAPABILITY_LABELS were two independently-written lists of the same vocabulary, the exact shape flagged elsewhere today (#653, and the Go-vocabulary/DB-CHECK drift m131 fixed).

Did not add a test comparing the two lists, since that would just relocate the second list. Instead CAPABILITY_LABELS is now the only place a capability name is spelled; KNOWN_CAPABILITIES and a new Capability type are derived from its keys (Object.keys / keyof typeof), so there is no second collection left to drift.

Mutation-tested both directions this turn, then reverted:
- Added a key to CAPABILITY_LABELS only, without touching KNOWN_CAPABILITIES, and confirmed via a throwaway script that it appeared in KNOWN_CAPABILITIES and capabilityLabel automatically.
- Assigned an unlisted string literal to the Capability type with no cast and got a real tsc error: TS2322, "not assignable to type" listing the known members.

So the "add it in one place, forget the other" mutation is no longer expressible, not just untested. capabilityLabel still takes a bare string (not Capability) and falls back to the raw wire value for anything unrecognised, since capabilitiesFromColumn in policy.go does not filter the column on read and a live grant can hold a name outside this build's vocabulary.

Gates rerun after the change: pnpm -C apps/web typecheck clean, pnpm -C apps/web lint 0 errors, vitest run src/features/ai-connections 186/186 passed, impeccable detect apps/web/src/features/ai-connections exit 0.

PR #668, opened 2026-09-01
feat(mcp): typed partial-failure envelope; stop disclosing tenant size to a scoped caller

mosamlife:
Adds the typed per-site partial envelope every fleet-wide tool needs, renames `list_sites` to `fleet_sites_list`, and fixes a live tenancy disclosure found while wiring it.

**Draft: layer 3 is a tenancy boundary and needs a `security-reviewer` pass.**

## The envelope

`asked` / `ok` / `refused` / `refusals[]`, in `apps/api/internal/mcp/envelope.go`.

`RefusalCode` is a struct with an unexported field, so the constructible values are exactly the package-level vars declared for it. **`site_out_of_scope` is declared nowhere and is unconstructible — a compile error, not a test.** `UnmarshalText` resolves against the same closed set, because without it `encoding/json` would populate the unexported field from arbitrary bytes and undo the guarantee at the deserialisation boundary.

## The counting decision

`asked` counts **the caller's own resolved scope**, never the tenant. `ResolveScopeSites` resolves every scope mode — including `all` — into an explicit id list joined through `sites` under RLS, so `auth.Sites.Len()` is exactly how many sites the connection may read.

`NewEnvelope` refuses any envelope where `ok + refused != asked`. The balance check is the leak detector: a residual is a site counted in `asked` and then dropped, which is what an out-of-scope site would leave.

## The leak this fixes

`ListSitesForModel` read a **tenant-wide** page (`sitesPageBound = 500`) and passed the page-bound flag into the rendered result — a fact about the tenant's row count, computed *before* the grant's site scope was applied. A connection scoped to 2 sites in a 600-site tenant received its 2 sites plus `"truncated": true` and a banner saying further sites were withheld: false for that caller, and enough to infer the tenant holds more than 500 sites.

The flag is no longer passed to the renderer. Completeness is recomputed over the caller's own scope, and an in-scope site the page did not return becomes an explicit `site_unread` refusal rather than a silent omission.

## Proofs

**Mutation M1 — layer 3 made to report instead of hide.** The mutation reports out-of-scope sites as refusals *and* counts `asked` over the tenant, so it still satisfies the balance check — only the hiding assertions can catch it.

```
# MUTATED
--- FAIL: TestMCPLayer3OutOfScopeSiteIsAbsentAndNotInferableAsAppRole (1.93s)
    mcp_layer3_hiding_integration_test.go:167: LAYER 3 LEAK: out-of-scope site id
    28f99da0-3547-4c6f-a7cf-da459873c2ea appears in the response body
--- FAIL: TestMCPPageBoundDoesNotDiscloseTenantSizeAsAppRole (1.48s)

# RESTORED
--- PASS: TestMCPLayer3OutOfScopeSiteIsAbsentAndNotInferableAsAppRole (1.99s)
--- PASS: TestMCPPageBoundDoesNotDiscloseTenantSizeAsAppRole (1.52s)
ok  github.com/mosamlife/wpmgr/apps/api/tests  4.327s
```

`git grep -q "MUTATION M" -- apps/api` exits **1**; `git status --short` is empty.

**The A/B proof** runs through the mounted transport with a real bearer, as `wpmgr_app`, printing from inside the transaction under test:

```
InTenantTx (fleet_sites_list read path): current_user=wpmgr_app rolsuper=false rolbypassrls=false
the tenant transaction sees 2 sites; the grant is scoped to 1
layer 3 holds: asked=1 ok=1 refused=0, and site B is absent from 1028 bytes of body
```

Both sites are in the **same tenant** deliberately: cross-tenant isolation is already proved by RLS elsewhere, so a two-tenant fixture would have RLS doing the work and would pass with the scope filter deleted. The test asserts the repo sees both rows first, then asserts exact numbers (`asked=1`, not a shape) and absence over the **raw response body**.

## Checks run

- `go build ./...`, `go vet ./...` — clean
- `go test ./internal/... ./cmd/...` — no failures
- `make test-integration` equivalent, full: `ok apps/api/tests 970.703s`, plus contract/gh458/testinfra all `ok`, **zero FAIL lines**. 545 integration tests (`grep -rhoE '^func Test[A-Za-z0-9_]+' apps/api/tests/*.go | wc -l`).

## Also fixed here

Near-miss name tests in both `registry_test.go` and `adr064_s7_...` spelled variants of `list_sites`, so after the rename they had stopped probing the live name while still passing. They now spell near misses of `fleet_sites_list`, and the retired name is kept as its own case — after a rename the old name is the likeliest string a client sends, and it must refuse rather than alias.

The tenant-cardinality guard was first written as a substring search and **reddened a correct result on its first run** (small integers occur inside uuids and `byte_cap:30720`). It now walks decoded JSON numbers by field name.

## Layer 7

Deferred, and said so in code. Phase 1 exposes no write tool, so there is no execution path for a late re-check to guard; the first write tool must bring it. A stub would have read as done.

## Summary by CodeRabbit

* **New Features**
  * Renamed the site-listing MCP tool to `fleet_sites_list`.
  * Added structured results that distinguish successful sites from sites that could not be read.
  * Added clear refusal details for stale, unavailable, or otherwise inaccessible sites.
  * Results now reflect only the caller’s permitted site scope, preventing disclosure of unrelated site counts.

* **Bug Fixes**
  * Improved handling of bounded and partially unavailable site listings.
  * Preserved clear truncation indicators when responses exceed limits.

### DIFF
--- a/apps/api/internal/mcp/audit_fail_closed_test.go
+++ b/apps/api/internal/mcp/audit_fail_closed_test.go
@@ -80,7 +80,7 @@ func routerWithRecorder(t *testing.T, store Store, rec auditRecorder) *gin.Engin
 	return r
 }
 
-const toolCallBody = `{"jsonrpc":"2.0","id":9,"method":"tools/call","params":{"name":"list_sites","arguments":{}}}`
+const toolCallBody = `{"jsonrpc":"2.0","id":9,"method":"tools/call","params":{"name":"fleet_sites_list","arguments":{}}}`
 
 // TestToolCall_AnswerIsWithheldWhenTheAuditAppendFails is the red-first half:
 // the append fails, and the tool's answer must not reach the client.
@@ -140,8 +140,8 @@ func TestToolCall_HealthyAuditStillAnswersAndWritesExactlyOneRow(t *testing.T) {
 	// only() fails loudly when there is no row, so "recorded nothing" cannot
 	// pass as "recorded correctly".
 	e := rec.only(t, audit.ActionMCPToolCalled)
-	if e.TargetID != ToolListSites {
-		t.Errorf("recorded target = %q, want %q", e.TargetID, ToolListSites)
+	if e.TargetID != ToolFleetSitesList {
+		t.Errorf("recorded target = %q, want %q", e.TargetID, ToolFleetSitesList)
 	}
 }
 
@@ -182,7 +182,7 @@ func TestRecordToolCall_PropagatesTheAppendFailure(t *testing.T) {
 	svc := NewService(&fakeStore{}).withAuditRecorder(&failingRecorder{err: boom})
 
 	auth := authWith(CapabilitySet{}, uuid.New())
-	err := svc.RecordToolCall(t.Context(), auth, ToolListSites, "read")
+	err := svc.RecordToolCall(t.Context(), auth, ToolFleetSitesList, "read")
 	if err == nil {
 		t.Fatal("RecordToolCall swallowed a failed append; the fail-closed gate above it " +
 			"can never fire")
@@ -198,7 +198,7 @@ func TestRecordToolCall_RecorderlessIsAnErrorNotASkip(t *testing.T) {
 	svc := NewService(&fakeStore{})
 
 	auth := authWith(CapabilitySet{}, uuid.New())
-	if err := svc.RecordToolCall(t.Context(), auth, ToolListSites, "read"); err == nil {
+	if err := svc.RecordToolCall(t.Context(), auth, ToolFleetSitesList, "read"); err == nil {
 		t.Fatal("a Service with no recorder reported success for a row it did not write")
 	}
 }
--- a/apps/api/internal/mcp/connection_status_race_test.go
+++ b/apps/api/internal/mcp/connection_status_race_test.go
@@ -435,7 +435,7 @@ func TestConnectionStatusSnapshot_ConcurrentInitializeCannotProduceTheImpossible
 			injectErr = fmt.Errorf("record connect: %w", err)
 			return
 		}
-		if err := svc.RecordToolCall(ctx, auth, "list_sites", "read"); err != nil {
+		if err := svc.RecordToolCall(ctx, auth, "fleet_sites_list", "read"); err != nil {
 			injectErr = fmt.Errorf("record tool call: %w", err)
 		}
 	}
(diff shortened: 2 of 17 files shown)


mosamlife:
## F1 fixed — the regression test now crosses its own bound

The reviewer was right: the test was vacuous. It called `mcpRepo.ListSitesForRead(ctx, tenant, 3)` as a standalone pre-check, then invoked the tool, which used the hard-coded `sitesPageBound = 500`. Six sites never came near it, so every assertion held with the disclosure fully restored.

`pageBound` is now a `Service` field defaulted by `NewService`, with a test-only `WithPageBound`. The proof injects 3, so the production read genuinely crosses its bound and `more` is true inside `ListSitesForModel`. `WithPageBound` panics on a non-positive bound rather than normalising to 500 — silently widening it would restore the exact vacuous pass it exists to prevent.

**MUTATION M2 — the shipped 0.61.152 disclosure restored verbatim:**

```
# MUTATED
--- FAIL: TestMCPPageBoundDoesNotDiscloseTenantSizeAsAppRole (1.95s)
    DISCLOSURE REGRESSION: truncation.truncated is true for a caller that
    received every site it may read. This is the tenant's page bound leaking
    through the scope filter, which is the defect this test exists for

# RESTORED
--- PASS: TestMCPLayer3OutOfScopeSiteIsAbsentAndNotInferableAsAppRole (1.90s)
--- PASS: TestMCPPageBoundDoesNotDiscloseTenantSizeAsAppRole (1.54s)
    page bound did not disclose tenant size: asked=1 truncated=false available=1
ok  github.com/mosamlife/wpmgr/apps/api/tests  4.551s
```

`git grep -q "MUTATION M" -- apps/api` exits **1**; `git status --short` is empty.

Header prose corrected: it claimed the role is asserted from inside the transaction the read uses. It is asserted inside a *separate* `InTenantTx` on the same pool.

## F2 — NOT fixed here; it is `database-engineer`'s path

Pushing the scope into SQL is the right fix and I am not making it. `ListSites` lives at `apps/api/db/query/sites.sql:27` and is **shared with the web `/sites` path** — it carries state, client_id, any_tags/all_tags, free-text `q` and six sort orders. Adding a scope predicate means either a new `sqlc.narg('site_ids')` arm on that shared query or a new dedicated `ListSitesForMCPScope`, plus the `internal/db/sqlc/**` regeneration that follows. Both are `apps/api/db/query/**`, which is not mine to write.

No migration is needed — `sites.id` is the PK and already indexed — so this is a query + regenerate, not a schema change.

**The defect stands as the reviewer described it**, and the second half is the worse one: the order is deterministic, so `site_unread` tells a caller scoped to a late-sorting site in a ≥500-site tenant to *retry*, and retrying can never help. It also implies a cardinality lower bound of 500. Not a regression — the old code dropped the site silently — but it is that bug wearing a new label, and the `site_unread` message currently promises a recovery that does not exist.

Issue #636, opened 2026-08-31
tools/call is served without a recorded initialize, so the wizard can report a working connection as never connected


mosamlife:
## Summary

`tools/call` and `tools/list` are dispatched without any check that the client ever sent `initialize`. The dispatch is a plain switch:

```go
// internal/mcp/transport.go, dispatch()
switch req.Method {
case "initialize":                      ...
case "notifications/initialized", "ping": ...
case "tools/list":                      ...
case "tools/call":                      ...
```

and nothing anywhere in the request path reads `client_identity_recorded_at` before serving. A client holding a valid bearer token can skip the handshake entirely and invoke tools.

## Two consequences, and the second is the reason to fix it

**1. It is a protocol violation we accept.** The transport specification treats `initialize` as the first request of a session, carrying the protocol revision and the client's declared capabilities. Serving tool calls to a client that never negotiated means the revision we think we agreed on was never agreed, and the capability exchange never happened.

**2. It makes the connection-status endpoint tell the truth and still be wrong.** `RecordClientIdentity` stamps `client_identity_recorded_at` during `initialize`. A client that skips it writes an `mcp.tool.called` audit row while that column stays NULL — so the wizard's status read reports `handshake.state="awaiting_client"` beside `first_call.state="succeeded"`.

That pair was treated as a race and fixed as one: the two reads were reordered so the impossible combination is unreachable by interleaving (`connection_status.go`, `ConnectionStatusSnapshot`). **That fix is correct and does not address this.** With no initialize gate, the pair is a faithful rendering of what is stored, and no amount of snapshot discipline changes it.

So the wizard can sit on "waiting for the client to connect" while the client is demonstrably working — the exact failure the step exists to prevent, arrived by the opposite route.

## What would close it

Two shapes, and they are not equivalent:

**A transport gate.** Refuse `tools/list` and `tools/call` before a recorded `initialize`, with a typed refusal naming the reason. This is what the specification implies and it makes the status endpoint's three states exhaustive. It is also a behaviour change on a deployed surface: any client currently skipping the handshake starts failing, which is the point, but it should be checked against the client survey before shipping.

**An explicit fourth state.** Accept that a client may skip the handshake and give the status read a state for it, so the wizard says something true rather than something contradictory. Cheaper, changes no behaviour, and leaves the protocol violation in place.

The first is better if no real client depends on the current leniency. That is a question the per-client survey in the wireframes can answer and nobody has asked it.

## Not established

[...]

Issue #631, opened 2026-08-31
impeccable detect exits 0 on an unreadable path and on a planted violation, so the design gate proves nothing


mosamlife:
## Summary

`impeccable detect` exits 0 both when it cannot read its input and when the input contains a violation it should catch. Executed today at `5418198f`:

```
$ apps/web/node_modules/.bin/impeccable detect /tmp/definitely-not-a-dir-xyz
Warning: cannot access /tmp/definitely-not-a-dir-xyz
EXIT=0

$ # plant  style={{ color: "#ff0000" }}  into a real .tsx, then:
$ apps/web/node_modules/.bin/impeccable detect apps/web/src/features/mcp-consent/consent-screen.tsx
EXIT=0        (no findings reported)
```

The planted line was removed afterwards; `git diff` on that file is clean.

## Why this is worse than a missing check

It is not that the linter is weak. It is that **its exit code is indistinguishable between "scanned and clean", "scanned and found nothing it knows how to find", and "did not scan anything at all"**. A step whose success and failure look identical carries none of the information its presence implies.

Two consequences, both live:

1. **It is in the definition of done for frontend work, and it has been reported as passing.** Multiple changes this week recorded `impeccable detect ... exit 0` as evidence of a clean design scan. That evidence is void — not wrong about the code necessarily, but not evidence either way. Anything gated on it should be regarded as ungated.
2. **The fail-open path means a moved or renamed directory silently stops being scanned.** A path argument that no longer resolves produces a warning on stderr and a zero exit, so a refactor can remove the check from the pipeline without removing it from the command line.

This is the shape CLAUDE.md names as the project's signature defect — announcing success over its own errors — sitting in the tool intended to catch defects.

## What would close it

Two separable things, and the first matters more:

**Exit non-zero when the input cannot be read.** A path that does not resolve, an empty file set, or zero files matched should be a hard error, not a warning. The pattern already used here is `scripts/check-rls-cross-tenant.sh`, which refuses to report a clean bill of health from an extraction that produced nothing and says so in those terms.

**Then establish what it actually detects for TSX**, because the second result above suggests the answer may be "nothing". Before treating any of its output as meaningful, plant one known violation of each rule it claims to implement and confirm each is reported. A rule nobody has watched fire is not known to be implemented.

## Relationship to existing work

[...]

PR #677, opened 2026-09-01
feat(web): give the connection-verification endpoint a screen

mosamlife:
## The gap

`GET /api/v1/mcp/connections/:id/status` has been mounted and serving since `apps/api/internal/mcp/handler.go:184`. Its own route comment says it exists for "the add-connection wizard's Step 8 and Step 9". **No dashboard code called it.**

Confirmed before starting, not assumed:

```
$ grep -rn "fetch(" apps/web/src --include="*.ts" --include="*.tsx" | grep -v "\.test\." | grep ai-connections
apps/web/src/features/ai-connections/use-ai-connections.ts:215   (GET list)
apps/web/src/features/ai-connections/use-ai-connections.ts:305   (POST create)
apps/web/src/features/ai-connections/use-ai-connections.ts:352   (POST revoke)
```

Three requests, and none of them the status endpoint. "Is my connection working?" had no answer in the product.

## What this adds

- `connection-status-model.ts` - the pure verdict mapping, so the sentence chosen is testable without a render.
- `use-connection-status.ts` - the poll. Hand-written zod, quoted field by field from `connectionStatusResponse` (`connection_status.go:738`). Confirmed there is no generated client: `grep -rn "mcp" packages/openapi-client/src/generated/` returns nothing, so this follows the sibling hooks' house pattern.
- `connection-verify.tsx` - the screen. Skeleton, `PageError`, and the verdict.
- Reachable from the connections list behind a per-row **Check connection** disclosure, offered on revoked connections too, because "did anything ever use this before I killed it" is exactly the question asked after a leak.

## What it refuses to say

The wireframe draws diagnoses this endpoint cannot support. Each is left undrawn and the gap is stated in the copy, not only in a comment:

| The deck draws | The response holds | What the screen says |
|---|---|---|
| S5 X, "3 attempts from 198.51.100.24 presented a key we do not recognise" | `refusal` hard-coded `null` (`connection_status.go:751`) | that it cannot tell a refusal from a never-started client |
| S5 X, "Version unsupported, we ended the session" | same gap, same reason | nothing |
| S29-9 P, "17 answered, 3 stale, 2 unreachable" | `partial` always `null` | that we do not record how many sites the call reached |
| S5 S, "Sites reachable 3 of 61", "Approvals it can grant" | neither number is in the response | nothing |

A refused client writes nothing to `mcp_grants`, so it is byte-identical to one that was never started. Naming a cause we do not have sends the operator to check a firewall over a client that was never restarted.

## Three states, kept apart

- **Never connected** is the shipped default and is not a failure. Under five minutes it reads as waiting; past that it says nothing has arrived; past thirty days it escalates the tone, never the claim. Both thresholds are the deck's own numbers and live in the browser because `connection_status.go:224` deliberately refuses to own them.
- **Connected and healthy** names the tool actually called, the client's own reported version, and the negotiated revision.
- **Connected and something is wrong** is the reachable one: a revision this build no longer speaks. It is not a refusal, and it does not claim to be.

`last_used_at` is reported, never used to derive the verdict. `RecordActivity` stamps it from `tools/list`, which every client issues unprompted, so deriving success from it would show "it read your fleet" for a client that read nothing (`connection_status.go:258`).

## Wizard numbering

Built to the ten-step drawing, per the owner's ruling. The panel carries the deck's step titles **without their numbers**: the shipped rail has four steps whose fourth ("Set it up") is the deck's step 6, so printing "Step 8" beside it would put two contradictory numbers on one screen. The deck is not renumbered down to match the code.

## Second commit: the contradiction #636 puts on this screen

`7360a4a4`. GH #636 says `tools/call` is served without a recorded `initialize`, so a live, working connection can carry `client_identity_recorded_at IS NULL` while its tool calls sit in the audit log. The naive render of that response prints "Nothing has reached us from this connection" directly above "It read your fleet".

Both cannot be true, and the false half is the absence: a recorded call is a thing that happened, a null column is a thing that was not written. The screen resolves towards the evidence and reports the mismatch as a gap in our own recording, while still declining to name the client or its revision, because that part really is unknown.

This was fixed here rather than filed: it is customer-visible, and it is in a file this PR already touches.

## Proof

Three real defects planted, four tests went red naming what was missing:

```
$ pnpm -C apps/web test -- --run src/features/ai-connections/connection-verify.test.tsx
x never connected > states that it cannot tell a refusal from a never-started client
x never connected > never renders an attempt count or a source address
x connected, and something is wrong > renders the indeterminate scan as not knowing, never as 'no calls yet'
x connected, and something is wrong > does not read last_used_at as proof of a first call
Tests  4 failed | 2145 passed (2149)
```

Restored, byte-exact (`git status --short` clean), green again.

A fourth defect planted for the second commit, the naive "print the absence anyway":

```
x never connected > never says 'nothing has reached us' about a connection that has demonstrably been used
x never connected > treats last_used_at alone as enough to disprove 'nothing has reached us'
Tests  2 failed | 2150 passed (2152)
```

A third test pins the over-fire half, so the branch cannot swallow the real never-used case.

Gates, final state:

```
$ pnpm -C apps/web typecheck   -> exit 0
$ pnpm -C apps/web lint        -> 10 problems (0 errors, 10 warnings), all pre-existing
$ pnpm -C apps/web test        -> Test Files 168 passed, Tests 2152 passed
$ pnpm -C apps/web build       -> built in 5.37s; routeTree.gen.ts unchanged (no new route)
$ apps/web/node_modules/.bin/impeccable detect apps/web/src/features/ai-connections -> no output, exit 0
```

Two caveats on that block, both stated rather than glossed:

1. **The impeccable result is weak evidence, per GH #631** ("impeccable detect exits 0 on an unreadable path and on a planted violation, so the design gate proves nothing"). Its exit code means nothing. What I can say is that the tool is not silent here: run over `apps/web/src` it prints 1 pre-existing anti-pattern at `components/layout/sidebar.tsx:642`, so it is reading files and reporting. Over the touched directory it prints nothing.
2. **One flake observed.** On one full-suite run, `src/routes/_authed/ai/-index.test.tsx > asks the path the API actually mounts` failed at 1065ms. It passes in isolation at 416ms and passed on the next two full runs. This PR adds a button to that page's table but no request, so a timing flake under load is the likely cause. Recording it rather than dropping it.

## Findings, not fixed here

- **The wizard cannot resume after the step 7 authorize redirect.** It holds all state in React and persists nothing, so an operator who leaves for the consent screen returns to step 1. Filed as #678.
- **#636 is the live cause of the contradiction handled above.** The screen now degrades honestly, but the underlying recording gap is in `apps/api` and is not this PR's to close.

## Where the deck is silent

Stated rather than invented:

- The `indeterminate` first-call state has no frame anywhere in the deck. The server added a third state (the scan hit its bound without finding a row) after the wireframes were drawn. Rendered as "we cannot tell", which is the only true sentence available.
- A **revoked** connection's verification has no frame. `connection_status.go:305` says the endpoint answers 200 for one deliberately so the wizard can stop polling, so the panel says so and stops.
- The deck never says what happens when the response contradicts itself (a `recognised` protocol state carrying a null version, or a `succeeded` first call carrying no timestamp). Both are rendered as our failure to read the answer, matching the vocabulary `connection-model.ts` already established for the list.

## Summary by CodeRabbit

* **New Features**
  * Added connection verification for AI connections, including handshake and first-call status.
  * Added per-connection “Check connection” controls with manual retry and automatic status updates.
  * Displays clear outcomes for healthy, revoked, inactive, and recording-gap scenarios.
  * Reports connection details such as tool name, client name, and protocol revision.

* **Bug Fixes**
  * Identifies cases where a credential was used but the session was not recorded, avoiding incorrect client fault reports.

### DIFF
--- a/apps/web/src/features/ai-connections/connection-status-model.ts
+++ b/apps/web/src/features/ai-connections/connection-status-model.ts
@@ -220,11 +220,29 @@ export function verifyVerdict(wire: ConnectionStatusWire, now: Date): VerifyVerd
         : ageMs >= NOTHING_ARRIVED_AFTER_MS
           ? "silent"
           : "fresh";
+    // DEGRADE THE VERDICT, NEVER THE EVIDENCE.
+    //
+    // This reads the WIRE and not `firstCall`, and the difference is a defect
+    // that shipped here for one commit. `firstCallVerdict` correctly degrades a
+    // `succeeded` carrying no `called_at` to `unknown` -- a success without its
+    // timestamp is a response we cannot read, and rendering it as a success
+    // would put a blank where the proof goes. But a flag whose entire job is to
+    // notice a contradiction must consult the raw claim, not the conclusion
+    // already drawn from it. Reading `firstCall.kind === "succeeded"` meant that
+    // a wire saying "a call succeeded", with the timestamp missing and
+    // last_used_at null, produced contradictedByUse === false and printed
+    // "Nothing has reached us from this connection" -- over a response that
+    // records a call.
+    //
+    // That is the GH #636 shape one field across: the same absence-beats-
+    // evidence inversion this branch was added to prevent, reintroduced by
+    // sourcing it from a degraded value.
+    //
     // Either half is evidence of use. last_used_at is too weak to prove a READ
     // (tools/list stamps it), but it is plenty to disprove "nothing has ever
     // reached us" -- something set it.
     const contradictedByUse =
-      firstCall.kind === "succeeded" || wire.first_call.last_used_at !== null;
+      wire.first_call.state === "succeeded" || wire.first_call.last_used_at !== null;
     return {
       kind: "live",
       handshake: { kind: "never_arrived", phase, ageMs, contradictedByUse },
--- a/apps/web/src/features/ai-connections/connection-status-model.ts
+++ b/apps/web/src/features/ai-connections/connection-status-model.ts
@@ -1,0 +1,223 @@
+// The verdict the verification screen renders, derived from one status poll.
+//
+// WHY A PURE MODULE. The whole risk on this screen is a sentence that claims
+// more than the response supports, and a sentence is easiest to pin in a test
+// when the thing that chose it is a function rather than a render. Every
+// branch below is reachable from a fixture in connection-status-model.test.ts.
+//
+// THE ONE RULE. "Nothing has reached us" and "something reached us and went
+// wrong" are different facts with different fixes, and this endpoint can only
+// tell us the first. The frontend must not upgrade one into the other. See
+// apps/api/internal/mcp/connection_status.go:149 (HandshakeRefusal) for the
+// server's own statement of the gap: a client refused for speaking a protocol
+// revision below the floor writes NOTHING to mcp_grants, so it is byte
+// identical to a client that was never started. The wireframe (S5, X frame)
+// draws that refusal as a live diagnosis. WE CANNOT DRAW IT. Where the deck
+// asks for a cause we do not have, this model returns the absence and the
+// screen says so in words.
+
+import type { ConnectionStatusWire } from "./use-connection-status";
+
+/**
+ * How long a connection may sit unused before the screen stops saying "any
+ * second now" and starts saying "nothing has arrived".
+ *
+ * FIVE MINUTES, AND THE NUMBER IS THE WIREFRAME'S. S5's timeout frame is
+ * titled "No client has connected in 5 minutes". THE SERVER REFUSES TO OWN
+ * THIS THRESHOLD ON PURPOSE -- connection_status.go:224 says a not-yet and a
+ * has-not-happened-for-a-while "are the same server-side fact and they differ
+ * only in how long the operator has been staring at it, which is a clock the
+ * browser owns". So the browser owns it, here, once, named.
+ */
+export const NOTHING_ARRIVED_AFTER_MS = 5 * 60 * 1000;
+
+/**
+ * When an unused connection stops being a note and becomes something to act on.
+ *
+ * THIRTY DAYS, from S5's E frame: "At 30 days we escalate this from a note to a
+ * banner." Escalating the TONE, never the CLAIM: a 40-day-old unused credential
+ * is still only an unused credential, and this model does not start calling it
+ * broken because it got old.
+ */
+export const ESCALATE_UNUSED_AFTER_MS = 30 * 24 * 60 * 60 * 1000;
+
+/**
+ * What the client said about the protocol revision it speaks.
+ *
+ * Deliberately NOT reusing ProtocolHeader from ./connection-model. That type
+ * answers the list's question ("what did it report"); this one answers the
+ * wizard's ("does that leave anything working differently"), and it carries the
+ * assumed floor as its own field because the status endpoint sends
+ * `assumed` separately from `version` for exactly that reason
+ * (connection_status.go:180: writing the floor into Version "would print a
+ * header the client never sent").
+ */
+export type ProtocolNote =
+  | { readonly kind: "recognised"; readonly version: string }
+  // The client sent no header and the specification says to assume the floor.
+  // A SUCCESS STATE. connection_status.go:104 calls this out: "treated as the
+  // floor, deliberately not an error".
+  | { readonly kind: "assumed"; readonly assumed: string }
+  // A revision is stored that this build no longer speaks. Reachable only
+  // through history (a revision dropped from the supported window after it was
+  // recorded), never through a live refusal.
+  | { readonly kind: "unrecognised"; readonly version: string }
+  // The response disagreed with itself: a state that must carry a version,
+  // carrying none. OUR failure to read the answer, never a claim about the
+  // client. Same reasoning as ProtocolHeader's `unreadable` kind.
+  | { readonly kind: "unreadable" };
+
+/** Has a client ever opened a session with this credential. */
+export type HandshakeVerdict =
+  | {
+      readonly kind: "never_arrived";
+      /**
+       * How loudly to say it. NOT three different facts -- one fact, three
+       * ages. `fresh` is the wizard's live wait, `silent` is past the
+       * five-minute mark, `stale` is past thirty days.
+       */
+      readonly phase: "fresh" | "silent" | "stale";
+      readonly ageMs: number;
+    }
+  | {
+      readonly kind: "connected";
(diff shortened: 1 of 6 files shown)


mosamlife:
All six threads addressed in `a8af7356`. Five applied, one applied on the client half only.

**`connection-verify.tsx:302` — invalid observation time.** Valid, and the worst of the six. Every age-derived verdict is measured from that instant, so substituting the browser clock produced a confident "waiting" / "nothing has arrived" / "unused for 40 days" with nothing behind it, and all three read exactly like a correct answer. It renders as not knowing now, with three tests: two on the state, one pinning the over-fire half so the guard cannot swallow every good response.

**`connection-verify.tsx:356` — terminal polling.** Valid, and my construction was wrong at the root. Disabling the second observer left the first registered, and one enabled observer is all TanStack needs to keep an interval running. There is now a single observer and the decision is a `stopWhen` predicate inside `refetchInterval`, where the data it depends on already lives.

**`connection-verify.tsx:333` — unbounded polling.** Valid. Ten-minute budget, a terminal state that says the checking stopped and that nothing is wrong with the connection because of it, and a "Check again" that resets it. Ten minutes sits well past the five-minute mark at which the screen already gives its answer, so the budget can only expire on a screen that has already answered.

**`connection-verify.test.tsx:254` — vacuous test.** Valid; it asserted the verdict shape under a name about polling. It asserts `shouldKeepPolling` directly now.

**`use-connection-status.ts:145` — zero cadence.** Valid, and the code contradicted its own comment: the comment argued that a legitimate 0 and a missing field are different answers, and the branch `ms <= 0` then folded them together. Zero stops now; missing, non-finite and negative take the floor. Worth noting for the record that `statusPollAfterMs` is a constant `2000` at `connection_status.go:63`, so zero is not reachable from today's server. Applied anyway, because the reading of intent is right and a comment disagreeing with its own code is a defect on its own.

**`use-connection-status.ts:149` — browser cache.** Valid, and applied on **the client half only**. `cache: "no-store"` is now on the fetch. The other half you name, `Cache-Control: no-store` on the status response, is in `apps/api` and is **not** in this PR: that path belongs to another agent working on it right now, and a same-file collision is a worse outcome than a deferred header. **This needs an API-side follow-up** — the client directive stops this app's own re-use, and only the response header stops a shared or intermediary cache.

### Re-run of the mutation set

A terminal-state change is exactly the kind that silently breaks a neighbouring assertion, so all seven defects were planted again together, including two new ones for the polling:

```
$ pnpm -C apps/web test -- --run src/features/ai-connections/connection-verify.test.tsx
x never connected > states that it cannot tell a refusal from a never-started client
x never connected > never renders an attempt count or a source address
x never connected > never says 'nothing has reached us' about a connection that has demonstrably been used
x never connected > treats last_used_at alone as enough to disprove 'nothing has reached us'
x connected and healthy > stops polling once the first call has succeeded
x connected, and something is wrong > renders the indeterminate scan as not knowing, never as 'no calls yet'
x connected, and something is wrong > does not read last_used_at as proof of a first call
x nextPollInterval > stops on a terminal verdict rather than polling a settled panel forever
x nextPollInterval > stops when the server names a cadence of zero
x an unreadable observed_at > renders as not knowing, never as a verdict from the browser clock
x an unreadable observed_at > renders no verdict at all in that state
x revoked > stops polling a revoked connection rather than waiting forever
Tests  12 failed | 2149 passed (2161)
```

Note the fifth line: the previously vacuous test now fails under a mutant, which it could not do before.

Restored byte-exact (`git status --short` clean), then:

```
$ pnpm -C apps/web typecheck   -> exit 0
$ pnpm -C apps/web lint        -> 10 problems (0 errors, 10 warnings), all pre-existing
$ pnpm -C apps/web test        -> Test Files 168 passed, Tests 2161 passed
$ pnpm -C apps/web build       -> built in 6.42s; routeTree.gen.ts unchanged
$ impeccable detect .../ai-connections -> no output, exit 0 (weak evidence, see GH #631)
```

mosamlife:
Sixth thread fixed in `6e1b35b3`. **The reviewer is right, and I verified the reachability before applying rather than taking it on relay.**

At `connection-status-model.ts:227` the flag read `firstCall.kind`, which `firstCallVerdict` has already degraded:

```ts
case "succeeded":
  return f.called_at === null ? { kind: "unknown" } : { kind: "succeeded", ... };
```

So a wire of `state: "succeeded"`, `called_at: null`, `last_used_at: null` gave `contradictedByUse === false`, and the panel printed "Nothing has reached us from this connection" over a body that records a call.

**This is the same inversion as `7360a4a4`, one field across.** There the handshake column was null while calls sat in the audit log; here the wire says a call succeeded and the flag consults the conclusion drawn from it instead of the claim itself. The degradation is correct and stays: a success without its timestamp is a response we cannot read, and rendering it as a success would put a blank where the proof goes. What was wrong is a contradiction detector sourcing itself from a degraded value. **Degrade the verdict, never the evidence.** It reads `wire.first_call.state` now.

On reachability: a well-behaved server always pairs the two, since `connection_status.go:246` sets `CalledAt` only for `FirstCallSucceeded`. But the `unknown` branch exists precisely because this code already refuses to trust that pairing, and the whole point of the zod boundary is that a malformed response must not become a confident wrong claim on screen. A guard that anticipates the contradiction and then reads past it is worse than not having it.

Why it survived the first pass: every earlier `succeeded` fixture carried a `called_at`. The case did not exist in the set.

**Fires:** reverting that one expression to the reviewed form, the new test is the only failure, and it names what is missing.

```
$ pnpm -C apps/web test -- --run src/features/ai-connections/connection-verify.test.tsx
x never connected > does not claim nothing arrived when the wire records a succeeded call with no timestamp
Tests  1 failed | 2161 passed (2162)
```

**Does not over-fire:** the new test also asserts the first-call half still degrades honestly to `unknown`, so the fix cannot be satisfied by rendering the unreadable success as a success.

Full mutation set re-run afterwards, since this touched an expression two earlier mutants already target. All seven planted together, **13 red** (12 before, plus the new one):

```
Tests  13 failed | 2149 passed (2162)
```

Restored byte-exact (`git status --short` clean):

```
$ pnpm -C apps/web typecheck   -> exit 0
$ pnpm -C apps/web lint        -> 10 problems (0 errors, 10 warnings), all pre-existing
$ pnpm -C apps/web test        -> Test Files 168 passed, Tests 2162 passed
$ pnpm -C apps/web build       -> built in 5.35s; routeTree.gen.ts unchanged
```

All six threads now addressed.

PR #683, opened 2026-09-01
fix(mcp): require the same permission on both routes that create a connection

mosamlife:
## What

Two routes create the same object — an MCP connection: a grant carrying a capability set and an organisation-wide site scope — and they required different permissions.

- `POST /api/v1/mcp/connections` (`RegisterConnections`) has always been behind `authz.RequirePermission(authz.PermAPIKeyManage)`.
- `GET /api/v1/oauth/mcp/authorize` and `POST /api/v1/oauth/mcp/consent` (`Register`) carried the org-scope gate and nothing else.

**Both routes that create a connection now require the same permission**, so the trust tier is a property of the object rather than of the path a caller happened to take to it.

## How

`requireGrantPermission`, a sibling of the existing `requireOrgScope`, mounted on the same pair of routes.

It is deliberately *not* `authz.RequirePermission` itself. That middleware refuses through `httpx` in the house `{code, message}` envelope; these two routes answer in the RFC 6749 §5.2 `{error, error_description}` shape, and the consent screen (`apps/web/src/features/mcp-consent/use-consent.ts`) parses exactly that. A correct refusal in the wrong envelope reaches that screen as an unexplained server fault. `requireOrgScope` exists for the same reason and this mirrors it.

The decision is not re-derived: the gate asks `authz.PrincipalAllows`, the same resolution `authz.RequirePermission` performs, so a capability-scoped principal is held to its explicit capability set and never to its role (m120/#510), and a later change to the permission matrix moves both routes together. Reading `Role` off the principal here would be a second opinion, and a second opinion drifts.

`403` via `ErrCodeAccessDenied`, not `401`: the credential is valid and re-presenting it cannot help.

## Proof

`apps/api/internal/mcp/consent_permission_test.go`, driven through the real `Handler.Register` middleware chain rather than against the service.

Both arms are mutation-proven, and the detailed red/green output is recorded off-repo:

- **Refusal.** Unmounting the middleware reddens the refusal tests; each asserts the RFC 6749 envelope and not merely the status, and asserts that no grant is written.
- **Mechanism.** Swapping `authz.PrincipalAllows` for a role read reddens a dedicated test, and independently trips `TestNoDirectAllowsOutsideAuthz` in `internal/authz`.
- **Over-fire.** A principal that legitimately holds the permission still completes `/authorize` and still mints at `/consent`, with `CreateGrantWithCode` asserted to have been reached. A guard that refuses everyone is not a fix, so this arm passes both before and after by design.

## Also

Integration fixtures that drove these routes with a principal holding no role are updated to carry the role the connections route already asks for. Their assertions are unchanged — the fixtures adapt to the rule; the rule is not loosened to fit the fixtures.

## Checks

- `go build ./...`, `go vet ./internal/mcp/... ./tests/...`, `go test ./internal/... ./cmd/... -count=1` — clean.
- `make test-integration` — sequenced separately; see the comment on the PR.
- No schema, no migration, no contract change; nothing regenerated.


mosamlife:
Second commit (`12b34d36`) covers a review finding and one hardening item.

**Third fixture.** `mountOAuthLikeProduction` in `tests/adr064_s16_mcp_connections_integration_test.go` built a principal that does not hold `PermAPIKeyManage`, so the three tests that call `connectRealGrant` died in setup — `TestMCPRevokeCascadesToTokensAsAppRole`, `TestMCPConnectionsListThroughMountedRouteAsAppRole` and `TestMCPGrantsSiteScopePolicyIsLiveAndTheServiceRefusesOutLoud`. The last of those is the `mcp_grants_site_scope` RLS proof, and `ci.yml` does not run this package, so it would have parked silently. The helper now uses `adminPrincipal`, the fixture the connections routes in the same file already use. Assertions unchanged.

**Sweep.** Two helpers in `apps/api/tests/` mount `mcp.Handler.Register` (the `/authorize` + `/consent` pair): `mountOAuthLikeProduction` and the parameterised `mountLikeProduction`. `mountStatusLikeProduction` and `mountConnectionsLikeProduction` mount `RegisterConnections` only. Searching for the path constants `mcp.AuthorizePath` / `mcp.ConsentPath` as well as the literal paths — the literals alone are what hid this one — the files that reach these two routes are `adr064_s16`, `adr064_s6b3`, `mcp_consent_org_scope`, `mcp_audit_actor_attribution` and `mcp_audit_events`. The last two already carried a principal that holds the permission. No fourth.

**Hardening.** `requireGrantPermission` now asks `domain.IsSiteConstrained` itself rather than depending on `requireOrgScope` being mounted ahead of it. `authz.RequirePermission`, which the connections route runs, carries both halves in one middleware; this makes the two routes equivalent under any mount order. The refusal text is a shared constant, so which of the two middlewares fires first is not observable. The branch is unreachable through the real pair, so it is proven by driving the middleware alone — planted and watched fail (`200`, admitted) before the check was restored.

**Integration suite**, run unpiped from the repo root:

```
make test-integration
ok  	github.com/mosamlife/wpmgr/apps/api/tests	974.191s
ok  	github.com/mosamlife/wpmgr/apps/api/tests/contract	3.215s
ok  	github.com/mosamlife/wpmgr/apps/api/tests/gh458	7.535s
ok  	github.com/mosamlife/wpmgr/apps/api/tests/testinfra	1.202s
```

No recipe failure from `make`. The same suite on the previous commit failed exactly the three tests named above, which is what identified the fixture.

The three, re-run and reported with their own status:

```
go test -count=1 -v -run "TestMCPRevokeCascadesToTokensAsAppRole|TestMCPConnectionsListThroughMountedRouteAsAppRole|TestMCPGrantsSiteScopePolicyIsLiveAndTheServiceRefusesOutLoud" ./tests/
--- PASS: TestMCPRevokeCascadesToTokensAsAppRole (2.03s)
--- PASS: TestMCPConnectionsListThroughMountedRouteAsAppRole (1.76s)
--- PASS: TestMCPGrantsSiteScopePolicyIsLiveAndTheServiceRefusesOutLoud (1.54s)
ok  	github.com/mosamlife/wpmgr/apps/api/tests	6.576s
THREE_NAMED_EXIT=0
```

Each reports `current_user=wpmgr_app rolsuper=false rolbypassrls=false`, and the RLS proof still logs `POLICY ok: the RESTRICTIVE site-scope policy refused the read (0 rows)`.

`go build ./...`, `go vet ./internal/mcp/... ./tests/...`, `go test ./internal/... ./cmd/... -count=1` clean. `git status --porcelain` empty; `git grep "MUTATION M" -- apps` exits 1.

mosamlife:
Re: Greptile's P2, **Denial skips OAuth callback** — the finding is correct, and it is being taken as a follow-up rather than folded in here. Reasoning, so the deferral is on the record:

**It is not introduced by this PR.** `requireOrgScope` has always refused a site-constrained principal the same way, with `AbortWithStatusJSON` and no callback. This change makes a second class of principal reach that same existing refusal. The shape is pre-existing and this PR is consistent with it.

**Fixing it in middleware would be worse than the bug.** RFC 6749 §4.1.2.1 says to redirect `error=access_denied` with the original `state` to the registered `redirect_uri`. But at middleware time the `redirect_uri` has not yet been validated against the registered client — that happens inside the authorize handler. Redirecting from the middleware would mean redirecting to an attacker-supplied URI, i.e. trading a hanging client for an open redirect. The correct fix moves the permission decision inside the handler, after client and redirect validation, and applies to **both** middlewares rather than just the new one.

That is a restructure of the authorize path, not a line change, and it does not belong in a permission fix that is otherwise proven and ready. Filing it as its own issue covering `requireOrgScope` and `requireGrantPermission` together.

Merging this; the denial-callback work follows separately.

PR #719, opened 2026-09-07
fix(agent): contain page-cache paths to the cache root

wildfang:
> **Note from the maintainers:** this description was edited to remove reproduction detail while the fix is unmerged. The original text is retained privately and nothing about the change itself has been altered. See the comment below.

Three related problems in the page-cache path handling.

1. **Dot-segment strip was not idempotent.** The strip in both the drop-in and `CacheKey::normalizePath` ran a single pass, so certain inputs could survive sanitisation and key reads and writes outside the request's host bucket. Both copies now run the strip to a fixpoint, and stay behaviourally identical so read and write agree.

2. **Host handling diverged between the two sides.** The drop-in's host charset regex was more permissive than the write side's `sanitizeHost`. The drop-in now rejects the affected host shapes (→ `unknown-host` bypass), matching the write side.

3. **No physical containment.** Neither the serve path (`readfile`) nor `CacheWriter::atomicWrite` verified the resolved path actually lives under `cache/wpmgr`. Both now realpath-check against the cache root — the serve side treats an escaped path as a MISS (boots WordPress), the write side refuses. This is the backstop for planted symlinks and any future sanitiser gap; it costs one realpath per HIT (PHP caches the resolution), and a symlinked cache root itself keeps working because both sides of the comparison resolve through the same link.

`WPMGR_PAGE_CACHE_DROPIN_VERSION` bumped 0.45.1 → 0.46.0 so `DropinInstaller` transparently reinstalls the fixed drop-in on existing sites.

**Tests:** cross-bucket traversal, host escape, symlinked bucket, and fixpoint unit cases. Green: 44 cache tests; full agent suite OK; PHPStan level 8 clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cache path handling to prevent directory traversal attempts from accessing another host’s cached content.
  - Added safeguards against malformed host values escaping the cache area.
  - Prevented cached files and writes from being accessed through symlink-based paths outside the permitted cache location.
  - Requests involving unsafe or invalid paths now correctly result in a cache miss.

- **Tests**
  - Added regression coverage for traversal attempts, malformed hosts, and symlink-based cache escapes.


mosamlife:
Thanks for this — and for the report generally. The review found the underlying bug is real and reproduces exactly as you describe.

One housekeeping change: I edited the description above to remove the reproduction detail, because this PR is public and the fix is not merged yet. The description named the precondition and gave a generator specific enough to use directly, on a path that runs unauthenticated on every front-end request. I confirmed it works verbatim against current `main`. Nothing about your change was touched — only the prose, and the original is kept privately.

No criticism intended: your email said you were holding public disclosure, and this was clearly the ordinary instinct to write a PR description that explains itself. It just lands differently on a public repo when the patch is still in review. Same reasoning applies to the parts of #716 and #718 that name what is still open; I have left those, as they are materially less actionable.

On the review itself, so you are not waiting: #716 and #717 come out clean, and I will come back on #718, #719 and #720 with specifics — there are a couple of real issues, including one in #718 that I think changes its risk profile rather than just its implementation. Detail to follow.

PR #736, opened 2026-09-11
feat(db): m135 seats the first non-read MCP capability, mcp.cache.purge

mosamlife:
Migration and schema only. No Go. A `backend-architect` follows and depends on this landing first.

## What this does

Widens `mcp_grants_capabilities_vocabulary_check` from m131's eight reads to those eight plus `mcp.cache.purge`. It stays a **closed allowlist** — a literal `<@ ARRAY[...]` containment check, not a `mcp.%` pattern — so an unknown string is still refused at the database.

m131 DECISION 2: *"the next migration that adds a member not ending in `.read` is the one that needs the write review."* This is that migration, which is why it is on its own diff.

## Why `mcp.cache.purge` is first

The agent-side `cache_purge` command is Write / Idempotent. The retry converges, it destroys no authored state (a page cache is derived data), and the worst outcome is a slower page load. The boundary under review is read-versus-write, not safe-versus-unsafe, so crossing it with the mildest available member is the point rather than a loophole.

The suffix is `.purge`, not `.write` or `.propose`. `.propose` would be a lie — no approval step exists and none is wanted here. `.write` would be broader than the grant, which confers exactly one operation.

## What does not change

No grant is widened, nothing is backfilled. Containment is monotone, so every existing `{mcp.sites.read}` row still passes; the CHECK is a ceiling and raising a ceiling moves nothing under it. Re-added **without** `NOT VALID`, so the table is scanned rather than left with an unchecked row. The constraint name is unchanged — the parity test looks it up by name.

`sqlc generate` (v1.31.1) produced no diff: a CHECK constraint changes no column and no query.

## Ordinal collision — needs a merge-order decision

This is `20260906000000/m135`, **not** m134. Unmerged branch `pr675` already holds `20260905000000_m134_mcp_updates_propose_capability.sql`, which drops and re-adds **this same constraint** over its own literal list.

Taking a free ordinal avoids a filename clash and nothing else. If both ever apply to one database, the higher ordinal applies second and **its literal wins entirely** — m135 would drop `mcp.updates.propose`. If a grant already holds that member the table scan fails and the migration errors inside `main()` at boot; if none does, the member vanishes silently and the first symptom is a 23514 at INSERT on the wizard path days later.

Whichever branch merges **second** owes a reconciling migration at a fresh ordinal naming both writes. This file states the collision rather than resolving it unilaterally — `mcp.updates.propose` is a proposal-gated write over update runs and is a separate review.

## Handoff to `backend-architect`

`TestCapabilityVocabularyMatchesTheDatabaseCheckAsAppRole` **fails now, by design** — the DB holds nine members and Go holds eight. The red test is the handoff.

The follow-up must decouple `DefaultGrantCapabilities()` from `AllCapabilities()` **in the same commit** that widens `capabilityVocabulary`, or every newly minted grant is stamped with the power to purge a live site's cache as a side effect of a one-line map edit.


mosamlife:
Superseded by #738, which contained this migration (`3cedfdf7`) plus the Go half, and has merged as `0f84ca3d`.

Merging this one alone would have left `main` with the database knowing `mcp.cache.purge` while the Go vocabulary did not — and CI would have stayed green, because the parity test that catches exactly that lives in `apps/api/tests`, which `ci.yml` deliberately excludes as container-based integration. The two halves had to land together, so they did.

